PR #1, opened 2022-05-31
Implement stump with addition code

Davidson-Souza:
This PR imports the basic logic for stump, along with addition.  
## This code can:  
-  Create an empty Stump  
- Add elements and update roots  
- Rewind old tree    
## Still missing:
- Proofs
- Deletion

Some basic tests are also provided.


src/accumulator/stump.rs
```diff
@@ -0,0 +1,139 @@
+use super::types;
```
kcalvinalvin, line 1: We should stick to 4 spaces for indentation

src/accumulator/stump.rs
```diff
@@ -0,0 +1,139 @@
...
+  /// empty.
+  ///# Example
+  /// ```
+  ///   use rustreexo::accumulator::stump::Stump;
+  ///   let mut s = Stump::new();
+  ///   let utxos = vec![];
+  ///   let stxos = vec![];
+  ///   s.modify(&utxos, &stxos);
+  /// ```
+  pub fn modify(&mut self, utxos: &Vec<bitcoin_hashes::sha256::Hash>, _stxos: &Vec<bitcoin_hashes::sha256::Hash>) {
```
kcalvinalvin, line 29: Could we change to `self` instead of `&mut self` for all these functions? Stump doesn't need to be mutated.

src/accumulator/stump.rs
kcalvinalvin, line 1: There's trailing spaces in these lines that should be removed: 18, 41, 45, 47, 48, 69, 94, 120, 135.

src/accumulator/stump.rs
kcalvinalvin, line 29: Stump wouldn't be able to take in stxos for deletion because it doesn't know where each of the txos would be located in the tree. Stump should take in a vector of u64s.

src/accumulator/stump.rs
```diff
@@ -0,0 +1,139 @@
...
+    /* Hardcoded test cases generated here: https://go.dev/play/p/pODpvB9NXAZ */
+    let fingerprints = [
+                                    /* Leafs */
+                                    [0x6e, 0x34], [0x4b, 0xf5], [0xdb, 0xc1], [0x08, 0x4f],
+                                    [0xe5, 0x2d], [0xe7, 0x7b], [0x67, 0x58], [0xca, 0x35],
+                                    
+                                    /* ... Stump don't stores internal nodes ...*/
+
+                                    /* Root */
+                                    [0xb1, 0x51]
```
kcalvinalvin, line 124: Just comparing the roots only would be enough since the roots are the hashes of all the leaves anyways. Also, we could compare the entire root instead of the first two bytes. The root for this tree is `b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42`. Modified test to print out the entire root here:

https://go.dev/play/p/b2qYaDHvJWV

src/accumulator/stump.rs
```diff
@@ -0,0 +1,139 @@
...
+
+
+#[cfg(test)]
+mod test {
+  use std::vec;
+  use bitcoin_hashes::{sha256, Hash, HashEngine};
+  use super::Stump;
+
+  
+  #[test]
```
kcalvinalvin, line 95: We should add more unit tests here to ensure the correctness of the code. Perhaps run the tests as macros (something like this https://github.com/BurntSushi/fst/blob/master/src/raw/tests.rs#L111-L158)

kcalvinalvin (changes requested):

src/accumulator/stump.rs
kcalvinalvin: Don't stress too much on this for now, I'll also write up some tests here.

src/accumulator/stump.rs
Davidson-Souza: Regarding code style, I ran `rustfmt` into the code, so it now conforms  rust's code standards. I think it's ok now.

src/accumulator/stump.rs
Davidson-Souza: > Just comparing the roots only would be enough since the roots are the hashes of all the leaves anyways  
I made this for catching errors with how do we hash things, perhaps making a new test for leaf hashes?

src/accumulator/stump.rs
Davidson-Souza: Sure. Didn't put too much effort on that, since deletion isn't implemented yet, but 027fe2d fixes this by changing modify's signature.

src/accumulator/stump.rs
kcalvinalvin: We should run it on the other files as well.. Should add it to the CI

kcalvinalvin:
Overall, I think we should add a bit more tests to make sure it's working right and it should be ok :)

src/accumulator/stump.rs
Davidson-Souza: I see what you mean. It's actually a good point.  
If modify is a pure function, then stump doesn't need to be mutable, every time you update a stump you create a new one and assign to a new variable. You may even shadow out the older stump, so you can use the same name after modifying.  
I'll experiment with that idea a little.

src/accumulator/stump.rs
Davidson-Souza: > We should run it on the other files as well  

I didn't run for this commit because those files are out of scope for this PR. But 100% agree in running `rustfmt` for everything else.  

> Should add it to the CI  

AFAIK, this codebase has no CI, this is a nice TODO.

src/accumulator/stump.rs
Davidson-Souza: Just took a few minutes to try that out, I've got a macro (for now called `make_test_case`) which creates tests for any tree with just the elements and roots.  
Here's how you implement the provided tree.
```rust
    make_test_case!(            
        tree_with_6_elements,
        [0, 1, 2, 3, 4, 5, 6],
        [
            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386",
            "9eec588c41d87b16b0ee226cb38da3864f9537632321d8be855a73d5616dcc73",
            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6"
        ]
    );
```
This one creates a new test function for each scenario. It is possible to have just one function with all cases within it, in that case, the macros would look like this:
```rust
    make_test_case!(
        [0, 1, 2, 3, 4, 5, 6],
        [
            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386",
            "9eec588c41d87b16b0ee226cb38da3864f9537632321d8be855a73d5616dcc73",
            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6"
        ]
    );
```
Another thing that came to my mind is providing test cases as a json or xml, so the code just read those cases and test.

src/accumulator/stump.rs
kcalvinalvin: Do you wanna do that in this PR or in a follow up PR?

src/accumulator/stump.rs
kcalvinalvin: >make_test_case!(
        [0, 1, 2, 3, 4, 5, 6],
        [
            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386",
            "9eec588c41d87b16b0ee226cb38da3864f9537632321d8be855a73d5616dcc73",
            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6"
        ]
    );

This looks great :)

> Another thing that came to my mind is providing test cases as a json or xml, so the code just read those cases and test.

Yeah that's a good idea. We'd eventually want some test cases across different implementations so making some test cases in json would be good. Bitcoin Core has them https://github.com/bitcoin/bitcoin/blob/master/src/test/data/tx_invalid.json

Davidson-Souza:
@kcalvinalvin 50e8714 adds some new tests, and use the JSON file idea. Any thoughts on that?

Cargo.toml
```diff
@@ -8,3 +8,5 @@ edition = "2018"
 [dependencies]
 bitcoin = "0.23.0"
 bitcoin_hashes = "0.7.6"
+sha2 = "0.10.2"
+serde_json="1.0.81"
```
kcalvinalvin, line 12: We could have this under `[dev-dependencies]` instead of `[dependencies]`

kcalvinalvin:
> @kcalvinalvin [50e8714](https://github.com/mit-dci/rustreexo/commit/50e8714703519ac8c6f5b61ac384bb9cdbbdea25) adds some new tests, and use the JSON file idea. Any thoughts on that?

The tests good! Looks great as a starting off point.

Cargo.toml
Davidson-Souza: Agree for `serde_json`. `sha2` can't be dev dependence because it's used on main code. `bitcoin-hashes` doesn't support sha512_256 as used for utreexo (it took me a little while to figure that out).

kcalvinalvin:
@Davidson-Souza is there any other changes that you're going to push?

It'd be better if you could squash the commits. Otherwise I can squash merge.

PR #4, opened 2022-07-01
Deletion code

Davidson-Souza:
This PR finalizes the Modify function, making it possible to delete elements from a stump.  Now (I think) we can have a real utreexo client, that keep its state and verifies proof as requested.   
## Now, this code can

- Modify the current state, adding new items.
- Undo the current state
- Verify proofs against a given state.

Moreover, there are a handful of tests for different scenarios, though more should be added in the future. I'm testing with a lot of different root counts, deleting siblings, deleting a node's sibling and one of its children, deleting a root and passing the targets unsorted.

### DIFF
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -1,7 +1,7 @@
 use crate::accumulator::{stump::Stump, types, util};
 use bitcoin_hashes::sha256;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 /// A proof is a collection of hashes and positions. Each target position
 /// points to a leaf to be proven. Hashes are all
 /// hashes that can't be calculated from the data itself.
@@ -137,7 +137,114 @@ impl Proof {
         }
         Ok(true)
     }
+    pub fn proof_after_deletion(
+        &self,
+        num_leafs: u64,
+    ) -> Result<(Vec<bitcoin_hashes::sha256::Hash>, Proof), String> {
+        let total_rows = util::tree_rows(num_leafs);
+        let mut targets = self.targets.to_owned();
+        targets.sort();
 
+        let proof_positions = util::get_proof_positions(&targets, num_leafs, total_rows);
+
+        let mut proof_hashes: Vec<_> = proof_positions
+            .to_owned()
+            .into_iter()
+            .zip(self.hashes.to_owned())
+            .collect();
+
+        let mut targets: Vec<_> = util::detwin(targets.to_owned(), total_rows)
+            .into_iter()
+            .rev()
+            .collect();
+
+        // Here is where our modified proof will be constructed
+        let mut target_hashes: Vec<(u64, bitcoin_hashes::sha256::Hash)> = vec![];
+
+        // For each of the targets, we'll try to find the sibling in the proof hashes
+        // and promote it as the parent. If it's not in the proof hashes, we'll move
+        // the descendatns of the existing targets and proofs of the sibling's parent
+        // up by one row.
+
+        while let Some(target) = targets.pop() {
+            if util::is_root_position(target, num_leafs, total_rows) {
+                target_hashes.push((target, bitcoin_hashes::sha256::Hash::default()));
+                continue;
+            }
+
+            let sib = target ^ 1;
+            // Is the node sibling on hash proofs?
+            if let Some(idx) = proof_hashes.iter().position(|(pos, _)| sib == *pos) {
+                let parent_pos = util::parent(sib as u64, total_rows);
+
+                target_hashes.push((parent_pos, proof_hashes[idx].1));
+
+                // Delete the sibling from proof_hashes as this sibling is a target now, not a proof.
+                proof_hashes.remove(idx);
+            } else {
+                // If the sibling is not in the proof hashes or the targets,
+                // the descendants of the sibling will be moving up.
+                //
+                // 14
+                // |---------------\
+                // 12              13
+                // |-------\       |-------\
+                // 08      09      10      11
+                // |---\   |---\   |---\   |---\
+                // 00  01          04  05  06  07
+
+                let to_update: Vec<_> = target_hashes
+                    .iter_mut()
+                    .filter(|(node, _)| {
+                        let parent = util::parent(sib, total_rows);
+                        util::is_ancestor(parent, *node, total_rows).unwrap_or(false)
+                    })
+                    .collect();
+
+                for to_update_node in to_update {
+                    let new_pos = util::calc_next_pos(to_update_node.0, sib, total_rows)?;
+                    to_update_node.0 = new_pos;
+                }
+
+                let to_update: Vec<_> = proof_hashes
+                    .iter_mut()
+                    .filter(|(node, _)| {
+                        let parent = util::parent(sib, total_rows);
+                        util::is_ancestor(parent, *node, total_rows).unwrap_or(false)
+                    })
+                    .collect();
+
+                for node_to_update in to_update {
+                    let next_pos = util::calc_next_pos(node_to_update.0, sib, total_rows)?;
+                    node_to_update.0 = next_pos;
+                }
+            }
+        }
+        let mut new_proof_hashes = vec![];
+        let mut new_target_hashes = vec![];
+
+        proof_hashes.sort();
+        target_hashes.sort();
+
+        proof_hashes
+            .iter()
+            .for_each(|(_, hash)| new_proof_hashes.push(hash.to_owned()));
+        let mut prove_targets = vec![];
+        target_hashes.into_iter().for_each(|(pos, hash)| {
+            prove_targets.push(pos);
+            new_target_hashes.push(hash)
+        });
+
+        prove_targets.dedup();
+
+        Ok((
+            new_target_hashes,
+            Proof {
+                targets: prove_targets,
+                hashes: new_proof_hashes,
+            },
+        ))
+    }
     pub fn create_root_candidates(
         &self,
         del_hashes: &Vec<sha256::Hash>,
@@ -227,7 +334,7 @@ impl Proof {
 #[cfg(test)]
 mod tests {
     use core::panic;
-    use std::str::FromStr;
+    use std::{str::FromStr, vec};
 
     use bitcoin_hashes::{sha256, Hash, HashEngine};
 
@@ -296,6 +403,69 @@ mod tests {
         if let Ok(res) = p.verify(&del_hashes, &s) {
             assert!(expected == res);
         }
+    }
+    #[test]
+    fn test_proof_after_deletion() {
+        let mut hashes = vec![];
+        for i in 0..20 {
+            hashes.push(hash_from_u8(i as u8));
+        }
+        let mut s = Stump::new();
+        s.modify(&hashes, &vec![]);
+
+        let proof = Proof::new(
+            vec![1, 16, 10],
+            vec![
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "e7cf46a078fed4fafd0b5e3aff144802b853f8ae459a4f0c14add3314b7cc3a6",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "4a64a107f0cb32536e5bce6c98c393db21cca7f4ea187ba8c4dca8b51d4ea80a",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "cd9c77062a338e63a63ca623db438cb8676f15466641079ee61ec2dda98de796",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "96d56447466674521007145ed72f8757517c72f7737dc4a0dcd3ecb996968971",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7",
+                )
+                .unwrap(),
+                bitcoin_hashes::sha256::Hash::from_str(
+                    "e799acb98a071c4884707e4bc8c093ba22571c8d84cc0223ab0c2c9327313a5b",
+                )
+                .unwrap(),
+            ],
+        );
+        let proof_after = proof.proof_after_deletion(s.leafs).unwrap();
+        let (del_hashes, proof) = proof_after;
+        let roots = proof.create_root_candidates(&del_hashes, &s).unwrap();
+
+        // They are swapped, because create_root_candidates builds the forest botton-up
+        let expected = vec![
+            bitcoin_hashes::sha256::Hash::from_str(
+                "21326d8aebeb6ef7bc02f40bdf778a02ba1c836b257f946ae21cab2a6f95fa18",
+            )
+            .unwrap(),
+            bitcoin_hashes::sha256::Hash::from_str(
+                "37968ef73d30dda38ede8357d66593c72acd4f0eb9f7a1a9acfeb7de850c05b4",
+            )
+            .unwrap(),
+        ];
+        assert_eq!(expected, roots);
     }
     #[test]
     fn test_proof_verify() {
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -62,7 +62,7 @@ impl Proof {
     ///   let mut proof_hashes = Vec::new();
     ///   let targets = vec![0];
     ///   // For proving 0, we need 01, 09 and 13's hashes. 00, 08, 12 and 14 can be calculated
-    ///   //Fill `proof_hases` up with all hashes
+    ///   //Fill `proof_hashes` up with all hashes
     ///   Proof::new(targets, proof_hashes);
     /// ```  
     pub fn new(targets: Vec<u64>, hashes: Vec<sha256::Hash>) -> Self {
@@ -79,7 +79,7 @@ impl Proof {
     ///   use bitcoin_hashes::{sha256, Hash, HashEngine};
     ///   use std::str::FromStr;
     ///   use rustreexo::accumulator::{stump::Stump, proof::Proof};
-    ///   let mut s = Stump::new();
+    ///   let s = Stump::new();
     ///   // Creates a tree with those values as leafs
     ///   let test_values:Vec<u8> = vec![0, 1, 2, 3, 4, 5, 6, 7];
     ///   // Targets are nodes witch we intend to prove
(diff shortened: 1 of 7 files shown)


Davidson-Souza:
This is a huge diff, isn't it :smile:.  
> Can we put get_proof_postions() into a separate PR or a separate commit?  

I'll try make a new PR with only `get_proof_positions`.

kcalvinalvin:
Could you rebase this on top of main?

Davidson-Souza:
> Could you rebase this on top of main?  

Done!

test_values/test_cases.json
```diff
@@ -252,5 +252,196 @@
...
+      ],
+      "proof_hashes": [
+       "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+       "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+       "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+       "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b"
+      ]
+    },
+    {
+      "leaf_values": [
```
kcalvinalvin, line 285: Nit: leaf_values for the test case below have different formatting but they're the same thing. Having the same formatting would be good

test_values/test_cases.json
```diff
@@ -252,5 +252,196 @@
...
+      ],
+      "roots": [
+        "d6ab1f11dd0ceee1862aa84b5e4d2c7520a8651fa2652529e8c49cc2b43b4476",
+        "9c053db406c1a077112189469a3aca0573d3481bef09fa3d2eda3304d7d44be8",
+        "55d0a0ef8f5c25a9da266b36c0c5f4b31008ece82df2512c8966bddcc27a66a0",
+        "4d7b3ef7300acf70c892d8327db8272f54434adbc61a4e130a563cb59a0d0f47"
+      ]
+    },
+    {
+      "leaf_values": [
```
kcalvinalvin, line 404: Nit: Same for here. leaf_values for the test case below have different formatting but they're the same thing. Having the same formatting would be good

test_values/test_cases.json
```diff
@@ -252,5 +252,196 @@
       ],
       "expected": true
     }
+
```
kcalvinalvin, line 255: Could we order the fields in the same order? Some cases have `roots` before `proof_hashes` but some don't. It'd be easier to read if they were in the same order.

src/accumulator/util.rs
```diff
@@ -12,27 +12,62 @@ pub fn is_root_position(position: u64, num_leaves: u64, forest_rows: u8) -> bool
     return root_present && root_pos == position;
 }
 
-// extractTwins is a optimization for batched deletions. It checks if the nodes
-// being deleted also have their sibling being deleted. It returns the parents
-// of the deleted siblings along with nodes that didn't have a sibling
-pub fn extract_twins(nodes: Vec<u64>, forest_rows: u8) -> (Vec<u64>, Vec<u64>) {
-    let mut parents = Vec::new();
-    let mut twined = Vec::new();
+// removeBit removes the nth bit from the val passed in. For example, if the 2nd
```
kcalvinalvin, line 15: Could we move these util functions into a different commit or a new PR? Would be very helpful for reviewing.

Davidson-Souza:
@kcalvinalvin I broke things down in a few commits and made the tests' field naming more consistent. Please let me know what you think.

src/accumulator/util.rs
```diff
@@ -12,27 +12,62 @@ pub fn is_root_position(position: u64, num_leaves: u64, forest_rows: u8) -> bool
...
+    let lower = val & lower_mask;
+
+    ((upper >> 1) | lower) as u64
+}
+pub fn calc_next_pos(position: u64, del_pos: u64, forest_rows: u8) -> Result<u64, String> {
+    let del_row = detect_row(del_pos, forest_rows);
+    let pos_row = detect_row(position, forest_rows);
+
+    if del_row < pos_row {
+        return Err("calcNextPosition fail.".to_string());
```
kcalvinalvin, line 33: Should mention why we failed in the error message. Should mention that since the passed in `del_pos` is in a lower row than `position`, we've failed.

src/accumulator/util.rs
```diff
@@ -291,6 +359,14 @@ mod tests {
...
         assert!(super::is_right_sibling(0, 1));
     }
+
+    #[test]
+    fn test_remove_bit() {
+        // This should remove just one bit from the final number
+        // 15 = 1111, removing bit 3 makes it 111, that is 7
+        let res = super::remove_bit(11, 2);
+        assert_eq!(res, 7);
+    }
```
kcalvinalvin, line 369: It's easy to add more tests here.

Could you add some more? Just something formatted like

```
        // 1010 => 101
        let res = super::remove_bit(10, 0);         
        assert_eq!(res, 5);                                    

        // 1110 => 110                          
        let res = super::remove_bit(14, 1);
        assert_eq!(res, 6);
```

should be enough

src/accumulator/util.rs
```diff
@@ -12,27 +12,62 @@ pub fn is_root_position(position: u64, num_leaves: u64, forest_rows: u8) -> bool
     return root_present && root_pos == position;
 }
 
-// extractTwins is a optimization for batched deletions. It checks if the nodes
```
kcalvinalvin, line 15: For future reference:

Extract_twins is really being removed and then remove_bit, calc_next_pos, and detwin is being added. But if you do this in a single step, `git diff` thinks that extract_twins is being modified to be remove_bit.

It's fine here because this is a small commit but for bigger commits and for bigger projects, it'll be a lot more helpful during reviewing if you did these steps separately.

src/accumulator/util.rs
```diff
@@ -12,27 +12,62 @@ pub fn is_root_position(position: u64, num_leaves: u64, forest_rows: u8) -> bool
...
 
-    for n in node_iter {
+    // This is the bit to be prepended.
+    let to_row = pos_row + 1;
+    let higher_bits = (1 << to_row) << (forest_rows - to_row) as u64;
+
+    // Put the bits together and return it.
+    Ok(higher_bits | lower_bits)
+}
+pub fn detwin(nodes: Vec<u64>, forest_rows: u8) -> Vec<u64> {
```
kcalvinalvin, line 46: Could we add tests for detwin?

Something like:

```
    #[test]
    fn test_detwin() {
        // 14
        // |---------------\
        // 12              13
        // |-------\       |-------\                                          
        // 08      09      10      11       
        // |---\   |---\   |---\   |---\                                          
        // 00  01  02  03  04  05  06  07       
        let mut targets: Vec<u64> = vec![0, 1, 4, 5, 7];
        targets = super::detwin(targets, 3);        
        assert_eq!(targets, vec![7, 8, 10]);                   

        targets = vec![4, 6, 8, 9];             
        targets = super::detwin(targets, 3);
        assert_eq!(targets, vec![4, 6, 12]);
    } 
```

Would be good.

src/accumulator/util.rs
```diff
@@ -12,27 +12,62 @@ pub fn is_root_position(position: u64, num_leaves: u64, forest_rows: u8) -> bool
...
         // If the next node in line is the current node's sibling
         // grab the parent as well
-        if n[0] | 1 == n[1] {
-            parents.push(parent(n[0], forest_rows));
-            twined.push(n[0]);
-            twined.push(n[1]);
+        let i = dels_after[(n) as usize];
+        let j = dels_after[(n + 1) as usize];
+
+        if i | 1 == j {
```
kcalvinalvin, line 56: `s/i | 1 == j/is_right_sibling(i, j)/`

src/accumulator/util.rs
```diff
@@ -215,6 +249,35 @@ pub fn root_position(num_leaves: u64, row: u8, forest_rows: u8) -> u64 {
...
+    }
+    if rise > forest_rows {
+        return Err("Cannot rise more than the forestRows".to_string());
+    }
+    let mask = ((2 << forest_rows) - 1) as u64;
+    Ok((pos >> rise | (mask << (forest_rows - (rise - 1)) as u64)) & mask)
+}
+
+pub fn is_ancestor(higher_pos: u64, lower_pos: u64, forest_rows: u8) -> Result<bool, String> {
+    let lower_row = detect_row(lower_pos, forest_rows);
```
kcalvinalvin, line 264: Should add
```
if higherPos == lowerPos {
    return Ok(false);
}
```

src/accumulator/util.rs
```diff
@@ -215,6 +249,35 @@ pub fn root_position(num_leaves: u64, row: u8, forest_rows: u8) -> u64 {
...
+
+    // Prevent underflows by checking that the higherRow is not less
+    // than the lowerRow.
+    if higher_row < lower_row {
+        return Ok(false);
+    }
+    // Return false if we error out or the calculated ancestor doesn't
+    // match the higherPos.
+    let ancestor = parent_many(lower_pos, higher_row - lower_row, forest_rows)?;
+    if higher_pos != ancestor {
```
kcalvinalvin, line 275: We can just return `Ok(higher_pos == ancestor)`

src/accumulator/util.rs
```diff
@@ -344,19 +405,23 @@ mod tests {
     }
     #[test]
     fn test_get_proof_positions() {
-        let targets: Vec<u64> = vec![4, 5, 7, 8];
-        let num_leaves = 8;
-        let targets =
-            super::get_proof_positions(&targets, num_leaves, super::tree_rows(num_leaves));
+        use crate::accumulator::util::tree_rows;
```
kcalvinalvin, line 408: There's no need fro changing this test in this commit right?

src/accumulator/util.rs
```diff
@@ -344,19 +405,23 @@ mod tests {
...
-        assert_eq!(vec![6, 9], targets);
+        assert_eq!(vec![1, 6, 9], hashes);
     }
     #[test]
     fn test_is_root_position() {
         let h = super::is_root_position(14, 8, 3);
         assert_eq!(h, true);
     }
     #[test]
+    fn test() {
```
kcalvinalvin, line 420: Should remove this test. I'm guessing it was used for debugging

kcalvinalvin:
Commit `9717beffb6bb41bc4e157ec1e01ca7156f4d3741` should be in a separate PR since it's not related to adding deletion to stump.

src/accumulator/util.rs
Davidson-Souza: I think this is a residual from changing `get_proof_positions`, I used to return a tuple, now it's just one vector. I think I changed the naming of the variables, and I also changed the test to something we discussed on #5. But this can be reverted to what we left in #5 without problem.

Davidson-Souza:
I think I addressed everything.  

> Commit 9717beffb6bb41bc4e157ec1e01ca7156f4d3741 should be in a separate PR since it's not related to adding deletion to stump.  

Revert them to the previous state for this pr?

kcalvinalvin:
> I think I addressed everything.
> 
> > Commit [9717bef](https://github.com/mit-dci/rustreexo/commit/9717beffb6bb41bc4e157ec1e01ca7156f4d3741) should be in a separate PR since it's not related to adding deletion to stump.
> 
> Revert them to the previous state for this pr?

Yeah just pop off that commit and make a new PR after merging this one

src/accumulator/proof.rs
```diff
@@ -137,7 +137,101 @@ impl Proof {
...
+                //
+                // 14
+                // |---------------\
+                // 12              13
+                // |-------\       |-------\
+                // 08      09      10      11
+                // |---\   |---\   |---\   |---\
+                // 00  01          04  05  06  07
+
+                let to_update: Vec<_> = target_hashes
```
kcalvinalvin, line 196: We should also be updating the `proof_hashes` here or else the function won't work correctly. I don't think the current tests don't catch that.

src/accumulator/util.rs
```diff
@@ -30,7 +30,7 @@ pub fn calc_next_pos(position: u64, del_pos: u64, forest_rows: u8) -> Result<u64
     let pos_row = detect_row(position, forest_rows);
 
     if del_row < pos_row {
-        return Err("calcNextPosition fail.".to_string());
+        return Err(format!("calc_next_pos faill: {del_row} < {pos_row}"));
```
kcalvinalvin, line 33: There's two `l`s here in fail

src/accumulator/util.rs
```diff
@@ -378,19 +385,12 @@ mod tests {
     }
     #[test]
     fn test_detwin() {
-        // 14
```
kcalvinalvin: What's the reasoning for removing the tree printf here?

src/accumulator/proof.rs
```diff
@@ -298,6 +392,70 @@ mod tests {
...
+                    "e799acb98a071c4884707e4bc8c093ba22571c8d84cc0223ab0c2c9327313a5b",
+                )
+                .unwrap(),
+            ],
+        );
+        let proof_after = proof.proof_after_deletion(s.leafs).unwrap();
+        let (del_hashes, proof) = proof_after;
+        let roots = proof.create_root_candidates(&del_hashes, &s).unwrap();
+
+        // They are swapped, because create_root_candidates builds the forest botton-up
```
kcalvinalvin, line 445: nit: typo here. Should be `bottom-up` with a m

src/accumulator/proof.rs
```diff
@@ -137,7 +137,101 @@ impl Proof {
...
+            .into_iter()
+            .rev()
+            .collect();
+
+        // Here is where our modified proof will be constructed
+        let mut target_hashes: Vec<(u64, bitcoin_hashes::sha256::Hash)> = vec![];
 
+        // For each of the targets, we'll try to find the sibling in the proof hashes
+        // and promote it as the parent. If it's not in the proof hashes, we'll move
+        // the descendatns of the existing targets and proofs of the sibling's parent
```
kcalvinalvin, line 166: nit: typo here. Should be `descendants`

kcalvinalvin (commented):
General nit: `leafs` should be `leaves`.

src/accumulator/proof.rs
Davidson-Souza: I changed this locally, but do you have test cases where it fails? It would be nice if we can test this before upstreaming.

Davidson-Souza:
> General nit: leafs should be leaves.  

This always confuses me, leaves sounds like to leave. `grep`-ing the code I found leafs almost everywhere :man_facepalming: 

> Also, it'd be easier for me to review if the new changes were committed to the relevant commits instead of as new commits  

Hmm, interesting. I thought that keeping the new changes on an always forward history would be better. By this, you mean, like squashing the changes onto the prior commits with interactive rebase?

kcalvinalvin:
> Hmm, interesting. I thought that keeping the new changes on an always forward history would be better. By this, you mean, like squashing the changes onto the prior commits with interactive rebase?

Yeah. It does depend on different organizations but Bitcoin Core and most Bitcoin related repositories will prefer that you squash the changes onto the prior commits with the interactive rebase

src/accumulator/proof.rs
kcalvinalvin: I took this from a failing fuzz test in the go library.

https://go.dev/play/p/ItlQAlQDxx3

There's two functions: one that updates the proof hashes and one that doesn't. The one that doesn't fails to verify

Davidson-Souza: The new commits fix it and adds this case in our tests. What do you think?

kcalvinalvin: Looks good :)

src/accumulator/util.rs
kcalvinalvin: I'd prefer changing this back to #5 before merging.

src/accumulator/util.rs
kcalvinalvin: With the change, we just get this as the error message.

```
`Err("calc_next_pos fail: 0 < 1")`'
```

Which is vague since the caller doesn't know what the 0 and 1 mean.

I suggest something like this:

```diff
diff --git a/src/accumulator/util.rs b/src/accumulator/util.rs
index 3504f8c..e43a4df 100644
--- a/src/accumulator/util.rs
+++ b/src/accumulator/util.rs
@@ -30,7 +30,7 @@ pub fn calc_next_pos(position: u64, del_pos: u64, forest_rows: u8) -> Result<u64
     let pos_row = detect_row(position, forest_rows);
 
     if del_row < pos_row {
-        return Err(format!("calc_next_pos fail: {del_row} < {pos_row}"));
+        return Err(format!("calc_next_pos fail: del_pos of {del_pos} is at a lower row than position at {position}"));
     }
 
     // This is the lower bits where we'll remove the nth bit.
```

test_values/test_cases.json
kcalvinalvin: The ordering is still mixed. I see that most of the test cases have the ordering of:

leaf_values, targets, roots, proof

but there's still many that don't follow the above ordering

kcalvinalvin:
Still some unaddressed reviews. Once those are fixed I think this can be merged.

Davidson-Souza:
Now I think everything got addressed. Should I squash?

kcalvinalvin:
Looks like you changed `leaf_values` to `leafs` in the latest push in file `test_cases.json`.

Davidson-Souza:
I think git is a bit confused about this naming because I removed that commit formatting this file. As we agreed on, that'll be a new PR. If I recall correctly, this is a proof test, those in those tests, I used `leafs` and not `leafs_values`. Looking at the commits, I couldn't even find where this changed.  Probably squashing should solve it.

kcalvinalvin:
> As we agreed on, that'll be a new PR

Ah ok. I was confused because I didn't see any new PRs open.

> Looking at the commits, I couldn't even find where this changed

This was a diff of the current commits and the commits before you force pushed. You've force pushed so this diff won't be in the current git history.

> Probably squashing should solve it.

I'd like to try to keep the commit history for this PR because I think commits `e9e07bdf94e356a2daad77b5cd239ed8463a799d`, `b5d140903489ef3f0c4c329ee581c129bcb16e41` and `
f5229c35cdeae30b05cec4d02d0a4ea058497103` are great since they each address a specific functionality. Squashing would make us lose this history which isn't ideal for the future where we might want to look at this commit by commit because there's a bug/would hurt people that like to read new code commit by commit.

Davidson-Souza:
I think it's all correct, but just give a second look at it because I'm not the best git user :laughing:.

kcalvinalvin:
One last thing:

On commit `9231fb3161f00088a7751c0691908a783e773f24` `src: add proof_after_deletion`, the build fails because of the doc since the changes to modify haven't been introduced yet. It builds with the following change.

```diff
diff --git a/src/accumulator/proof.rs b/src/accumulator/proof.rs
index cc4c08f..fc0778c 100644
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -79,7 +79,7 @@ impl Proof {
     ///   use bitcoin_hashes::{sha256, Hash, HashEngine};
     ///   use std::str::FromStr;
     ///   use rustreexo::accumulator::{stump::Stump, proof::Proof};
-    ///   let s = Stump::new();
+    ///   let mut s = Stump::new();
     ///   // Creates a tree with those values as leafs
     ///   let test_values:Vec<u8> = vec![0, 1, 2, 3, 4, 5, 6, 7];
     ///   // Targets are nodes witch we intend to prove
@@ -107,7 +107,7 @@ impl Proof {
     ///     hashes.push(hash);
     ///   }
     ///
-    ///   let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap();
+    ///   s.modify(&hashes, &vec![]);
     ///   let p = Proof::new(targets, proof_hashes);
     ///   assert!(p.verify(&vec![hashes[0]] , &s).expect("No error should happens"));
     ///```
```

Can we make this change here and apply ` `let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap();` code in `25361281a991a87615473cb66ba00399d35f6b8f`

Davidson-Souza:
Patched it, now I think they all compile on its own.

PR #5, opened 2022-07-07
Proof positions

Davidson-Souza:
This PR implements `get_proof_positions`, a function useful for batch deletion.

### DIFF
--- a/src/accumulator/util.rs
+++ b/src/accumulator/util.rs
@@ -235,6 +235,47 @@ fn next_pow2(n: u64) -> u64 {
     t |= t >> 32;
     t + 1
 }
+/// Returns whether a and b are sibling or not
+fn is_sibling(a: u64, b: u64) -> bool {
+    a ^ 1 == b
+}
+/// Returns which node should have its hashes on the proof, along with all nodes
+/// whose hashes will be calculated to reach a root
+pub fn get_proof_positions(targets: &Vec<u64>, num_leaves: u64, forest_rows: u8) -> Vec<u64> {
+    let mut proof_postions = vec![];
+    let mut computed_postions = targets.clone();
+
+    for row in 0..=forest_rows {
+        let mut row_targets = computed_postions
+            .to_owned()
+            .into_iter()
+            .filter(|x| super::util::detect_row(*x, forest_rows) == row)
+            .peekable();
+
+        while let Some(node) = row_targets.next() {
+            if is_root_position(node, num_leaves, forest_rows) {
+                let idx = computed_postions.iter().position(|x| node == *x).unwrap();
+
+                computed_postions.remove(idx);
+                continue;
+            }
+            if let Some(next) = row_targets.peek() {
+                if !is_sibling(node, *next) {
+                    proof_postions.push(node ^ 1);
+                } else {
+                    row_targets.next();
+                }
+            } else {
+                proof_postions.push(node ^ 1);
+            }
+
+            computed_postions.push(parent(node, forest_rows));
+            computed_postions.sort();
+        }
+    }
+
+    proof_postions
+}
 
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
Proofpositions fails on the below tree with targets: `4, 5, 7, 8`.

```
14:99a2                                                         
|-------------------------------\                               
12:dc6f                         13:e4e4                         
|---------------\               |---------------\               
08:0100         09:4cce         10:9cdb         11:ffc5         
|-------\       |-------\       |-------\       |-------\       
                02:0300 03:0400 04:0500 05:0600 06:0700 07:0800 
```

Should return proof positions of `6, 9` but instead the current code returns `10, 11, 12, 13`
Playground link to the example with the current go code: https://go.dev/play/p/jcNYaoi__ey

Also, is there a particular reason/advantage to returning the targets with the computable positions? In other implementations (C++ and Go) we're not including the targets in there.

kcalvinalvin:
Also, I think we can just have one commit instead of two. Would be preferable to squash the two together.

Davidson-Souza:
> Proofpositions fails on the below tree with targets: 4, 5, 7, 8.


Wait, I ran this:
```rust
let targets: Vec<u64> = vec![4, 5, 7, 8];
let num_leaves = 8;
let (targets, hashes) = super::get_proof_postions(&targets, num_leaves, tree_rows(num_leaves));

println!("{:?}", targets);
```
and it prints `[6, 9]`. What am I missing here?

> Also, is there a particular reason/advantage to returning the targets with the computable positions? In other implementations (C++ and Go) we're not including the targets in there.  

That's a good question, actually. I don't have any good reason for that, just look a bit natural when I coded this function, but this value is always dropped, perhaps this can be removed.  

> Also, I think we can just have one commit instead of two. Would be preferable to squash the two together.  

Ok, I can squash things, but I think there'll be more commits before merging.

kcalvinalvin:
> and it prints [6, 9]. What am I missing here?

Oh sorry looks like it does! My mistake 

> Ok, I can squash things, but I think there'll be more commits before merging.

Ok. Would these new commits be for tests?

Davidson-Souza:
> Ok. Would these new commits be for tests?  

I was thinking in some changes you request, but if you think this code is ok, we can wrap up.

kcalvinalvin:
> I was thinking in some changes you request, but if you think this code is ok, we can wrap up.

How about removing the targets from the computable positions to match the other implementations?

kcalvinalvin:
Oh I meant that we should be removing targets from the computable_positons as in:

We're returning `[4, 5, 7, 8, 10, 11, 12, 13]` but we should be returning `[10, 11, 12, 13]`.

Stump doesn't use computable_positions so not returning the computable_positions is ok but it should be edited later down the road if we were to support proof caching since I am using computable_positions for that.

kcalvinalvin:
`computed_postions.drain(0..targets.len()); ` won't work because there's no guarantee that the targets will be the lowest positions.

```
14:f907                                                         
|-------------------------------\                               
12:0100                         13:e4e4                         
|---------------\               |---------------\               
                                10:9cdb         11:ffc5         
|-------\       |-------\       |-------\       |-------\       
                                04:0500 05:0600 06:0700 07:0800 
```
For this tree, if the targets were `[6, 7, 12]`, then the computable should be `[11, 13]`.

Before calling `drain()`, variable `computed_positions` would be `[6, 7, 11, 12, 13]`. We'd be incorrectly removing 11 and keeping 12.
https://go.dev/play/p/cMajKEld_oF

The only simple way to remove the targets from `computed_positions` would be to iter through it and remove one by one.

I'm not opposed to just choosing not to return `computed_positions` for now since we won't be using them for the moment.

PR #8, opened 2022-09-21
Calculate hashes

Davidson-Souza:
Changes calculaet_roots to return all computed positions along with calculated roots.

### DIFF
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -266,7 +266,87 @@ mod tests {
 
         sha256::Hash::from_engine(engine)
     }
+    #[test]
+    fn test_calculate_hashes() {
+        // Tests if the calculated roots and nodes are correct.
+        // The values we use to get some hashes
+        let preimages = vec![0, 1, 2, 3, 4, 5, 6, 7];
+        let hashes = preimages
+            .into_iter()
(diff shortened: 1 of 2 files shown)


kcalvinalvin (commented):
Clever design! I guess the Rust builtins for vectors allow for this without much performance hit. I do wonder how the sort & push effects performance but we shouldn't pre-optimize so we can look at that later.

Since we're just returning the nodes used to calculate the roots the existing tests should cover the intermediate nodes as well.

Just the spelling error there and we should be good to go.

Davidson-Souza:
> Clever design! I guess the Rust builtins for vectors allow for this without much performance hit. I do wonder how the sort & push effects performance but we shouldn't pre-optimize so we can look at that later.  

Yeah, I'm testing with a priority queue, so you don't have to sort & push. But I think some basic benchmark is important to make any assertions about performance, so I'm doing that first.  

Forced-push 5cceb0c, fixes typo and rebases with main.

kcalvinalvin:
> Yeah, I'm testing with a priority queue, so you don't have to sort & push. But I think some basic benchmark is important to make any assertions about performance, so I'm doing that first.

You could also just not sort at all by implementing something like what's there in pytreexo.

PR #11, opened 2022-10-04
Rework tests from the json file

Davidson-Souza:
### (This PR may be broken in two, if requested)

The first commit makes the field naming similar to [pytreexo](https://www.github.com/utreexo/pytreexo). On the other hand, 0d20448 improves how our test code reads from `test_cases.json`. Instead of parsing json manually, I'm using a struct and serde_json to deserialize into a native struct, this makes the code cleaner and easier to read.

### DIFF
--- a/test_values/test_cases.json
+++ b/test_values/test_cases.json
@@ -1,22 +1,22 @@
 {
   "insertion_tests": [
     {
-      "leaf_values": [0, 1, 2, 3, 4, 5, 6, 7],
-      "roots": [
+      "leaf_preimages": [0, 1, 2, 3, 4, 5, 6, 7],
+      "expected_roots": [
         "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
       ]
     },
     {
-      "leaf_values": [0, 1, 2, 3, 4, 5, 6],
-      "roots": [
+      "leaf_preimages": [0, 1, 2, 3, 4, 5, 6],
+      "expected_roots": [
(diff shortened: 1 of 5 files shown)


test_values/test_cases.json
```diff
@@ -468,30 +468,30 @@
       ]
     },
     {
-      "leaf_values": [0, 1, 2, 3, 4, 5, 6, 7],
-      "target_values":[0, 1, 2, 3, 4, 5, 6, 7],
-      "roots":[
```
kcalvinalvin, line 436: Seems odd to me that these changes would be here when the name changes happened at the previous commit.

kcalvinalvin (commented):
I like the changes to the test file as that makes testing for non-perfect forests easier.

The change to serde json deserialize also makes sense to me as that gets rid of boilerplate code.

test_values/test_cases.json
Davidson-Souza: It's because of the struct fields, I may fix this if you think it's better.

kcalvinalvin: Could you explain by what you mean when you say it's because of the struct fields?

My reasoning for this was because all the `test_cases.json` changes seem to be in the previous commit but these were in the later commit.

Davidson-Souza: Hmm, I see. This field naming should change in the first commit. Now I think it's ok.

kcalvinalvin:
ACK 18a62a5df3407f9995561b368c1476cf3054d60a

Issue #12, opened 2022-10-09
Requirements for a node&wallet to use rustreexo library


kcalvinalvin:
For a node/wallet to used the rustreexo library, we need to support additional things in the current rustreexo library.

- [ ] add() method on stump needs to return new hashes and their positions
- [ ] del() method on stump needs to return new hashes in their current positions
- [ ] Need to support add() on cached proofs
- [ ] Need to support del() on cached proofs
- [ ] Extra: Need to support undo on cached proofs

Additionally, the node either needs to be able to parse utreexo block messages or the utreexod node in github.com/utreexo/utreexod needs to support an rpc call that returns utreexo proofs for a requested block.

PR #13, opened 2022-10-10
initial impl of UpdateData

Davidson-Souza:
This PR updates `modify's` api to also return what have changed during this update. For now, only roots changed during addition are returned. This is part of the steps in #12.

### DIFF
--- a/src/accumulator/stump.rs
+++ b/src/accumulator/stump.rs
@@ -1,4 +1,4 @@
-use super::{proof::Proof, types};
+use super::{proof::Proof, types, util};
 use bitcoin_hashes::sha256;
 use std::vec;
 
@@ -6,6 +6,15 @@ use std::vec;
 pub struct Stump {
     pub leafs: u64,
     pub roots: Vec<bitcoin_hashes::sha256::Hash>,
+}
+#[derive(Debug, Clone, Default)]
+pub struct UpdateData {
+    /// to_destroy is the positions of the empty roots removed after the add.
+    pub(crate) to_destroy: Vec<u64>,
+    /// pre_num_leaves is the numLeaves of the stump before the add.
+    pub(crate) prev_num_leaves: u64,
+    /// new_add are the new hashes for the newly created roots after the addition.
+    pub(crate) new_add: Vec<(u64, sha256::Hash)>,
 }
 
 impl Stump {
@@ -36,21 +45,23 @@ impl Stump {
     ///   let stxos = vec![];
     ///   let s = s.modify(&utxos, &stxos, &Proof::default());
     ///   assert!(s.is_ok());
-    ///   assert_eq!(s.unwrap().roots, utxos);
+    ///   assert_eq!(s.unwrap().0.roots, utxos);
     /// ```
     pub fn modify(
         &self,
         utxos: &Vec<bitcoin_hashes::sha256::Hash>,
         del_hashes: &Vec<bitcoin_hashes::sha256::Hash>,
         proof: &Proof,
-    ) -> Result<Stump, String> {
+    ) -> Result<(Stump, UpdateData), String> {
         let mut root_candidates = proof
             .calculate_hashes(del_hashes, self)?
             .1
             .into_iter()
             .rev()
             .peekable();
-        let mut computed_roots = self.remove(del_hashes, proof)?.into_iter().rev();
+
+        let (_, computed_roots) = self.remove(del_hashes, proof)?;
+        let mut computed_roots = computed_roots.into_iter().rev();
 
         let mut new_roots = vec![];
 
@@ -67,13 +78,19 @@ impl Stump {
 
             new_roots.push(*root);
         }
+        let (roots, updated, destroyed) = Stump::add(new_roots, utxos, self.leafs);
 
-        let roots = Stump::add(new_roots, utxos, self.leafs);
-
-        Ok(Stump {
+        let new_stump = Stump {
             leafs: self.leafs + utxos.len() as u64,
-            roots: roots,
-        })
+            roots,
+        };
+        let update_data = UpdateData {
+            new_add: updated,
+            prev_num_leaves: self.leafs,
+            to_destroy: destroyed,
+        };
+
+        Ok((new_stump, update_data))
     }
 
     /// Rewinds old tree state, this should be used in case of reorgs.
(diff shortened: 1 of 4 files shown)


src/accumulator/stump.rs
```diff
@@ -167,7 +193,36 @@ mod test {
...
 
+        let positions = vec![0, 1, 2, 3, 4, 5];
+
+        let hashes: Vec<_> = vec![
+            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
+            "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
+            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5",
+            "02242b37d8e851f1e86f46790298c7097df06893d6226b7c1453c213e91717de",
+            "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
```
kcalvinalvin, line 216: Just leaving a note:

The go library currently also returns the roots as well so if we don't return them here, the behaviors won't be the same.

Not sure if returning the root or not returning the root is better.

kcalvinalvin (commented):
Overall I like the changes. Would prefer more data driven test to ensure correctness of the code. At the moment I'm not quite so sure the code is correct.

src/accumulator/stump.rs
Davidson-Souza: Do you mean the current set of roots?

src/accumulator/stump.rs
```diff
@@ -100,53 +118,61 @@ impl Stump {
...
+
+            // You can say if a root is empty, by looking a the binary representations of the
+            // number of leafs. If the h'th bit is one, then this position is occupied, empty
+            // otherwise.
+            let mut to_add = add.clone();
+            let mut pos = leaves;
+            while (leaves >> h) & 1 == 1 {
+                let root = roots.pop();
+                if let Some(root) = root {
+                    updated_subtree.push((util::left_sibling(pos), root));
```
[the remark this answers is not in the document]
Davidson-Souza: Duplicated additions?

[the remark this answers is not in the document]
kcalvinalvin: Not duplicated additions but when you add multiple things, you calculate the same position multiple times.
For example, when you add 05 to the tree below, the "new" positions are `[05, 10]`.

```    
12                          12
|-------\                   |-------\
08      09           ==>    08      09      10
|---\   |---\               |---\   |---\   |---\
        02  03  04                  02  03  04  05
```         

In my implementation in go, I keep track of all the positions and their hashes used during the add.
For example, when you add `04` to the tree below, the used positions are `[04]`.
       
12                          12
|-------\                   |-------\
08      09           ==>    08      09
|---\   |---\               |---\   |---\
        02  03                      02  03  04
                    
Then when you add `05`, the used positions are `[04, 05, 10]`
        
```         
12                          12
|-------\                   |-------\
08      09           ==>    08      09      10
|---\   |---\               |---\   |---\   |---\
        02  03  04                  02  03  04  05
```
       
So when we add two elements to the below tree, we get:
                
`[04]`
`[04, 05, 10]`                          

Where there are two `04`s.
                 
```               
12                       
|-------\
08      09                   
|---\   |---\
        02  03
```                             
                
From what I can see in this PR, the process is mostly the same as my go implementation but I don't
see where the duplicates are handled. Thus I'm not certain if the code is correct.

src/accumulator/stump.rs
kcalvinalvin: Not the current set of roots. I mean the roots that were generated by the additions.

So here, the final root after adding all 6 elements was `df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386`.

In the go implementation, this root would also be returned.

Davidson-Souza:
Rebased with main, addressed all comments and added some new tests.

src/accumulator/stump.rs
```diff
@@ -100,53 +117,67 @@ impl Stump {
...
+                        to_add = types::parent_hash(&root, &to_add);
+                    }
+                }
+                h += 1;
+            }
+            updated_subtree.push((pos, to_add));
 
-        roots
-    }
+            updated_subtree.sort();
```
kcalvinalvin, line 174: In the go one I had a vector of subtrees because you could avoid the sort & dedupe here.

Maybe in rust those operations aren't as slow but just noting here for future reference.

test_values/cache_tests.json
```diff
@@ -0,0 +1,138 @@
...
+            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
+            "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
+            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5",
+            "02242b37d8e851f1e86f46790298c7097df06893d6226b7c1453c213e91717de",
+            "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
+            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+        ],
+        "to_destroy": []
+    },
+    {
```
kcalvinalvin, line 22: It'd be nice to know how you got to these roots. For others I can kinda guess but for this one I can't really tell. 

Maybe a "comment" field and add a simple description like "add 6, delete [0, 4]" or something similar would be nice.

kcalvinalvin:
ACK 360e7ab6fef98e49753cce7610940401bd646a1d

Verified the cache tests on the go library. The stump add algorithm looks ok to me.

Could you rebase on main again and address the comment I left on the tests?

Davidson-Souza:
> Could you rebase on main again and address the comment I left on the tests?

Done with [7d818ec](https://github.com/mit-dci/rustreexo/commit/7d818ec963bbd4caa00db13c014eb59b4c08f922)

test_values/cache_tests.json
```diff
@@ -0,0 +1,142 @@
...
+        "initial_stump": "Add six leaves [0, 1, 2, 3, 4, 5] and remove them",
```
kcalvinalvin, line 23: Could you specify which ones you removed here? It's not possible to have two none empty roots here if you remove all six leaves

Davidson-Souza: Ops, my mistake. This one has no deletions. Fixed!

kcalvinalvin:
ACK a337fa4c488e816e083d9b17fb2a1d4ccd43651f

PR #14, opened 2022-10-19
Remove test_macros module

kcalvinalvin:
The test_macros module being exported made the macros show up in the cargo documentation. Hiding it makes the documenation more readable and since the macros are only used in stump tests, it is moved to the tests.


kcalvinalvin:
Pinging @Davidson-Souza for review.

Davidson-Souza:
This module and its macros can be removed after #11. They are only useful for manual parsing of JSON, but with serde we can do this automatically. Do you see other usages of it outside tests?

kcalvinalvin:
> This module and its macros can be removed after https://github.com/mit-dci/rustreexo/pull/11. They are only useful for manual parsing of JSON, but with serde we can do this automatically.

Ok yeah checked that these macros aren't called in that branch.

> Do you see other usages of it outside tests?

Not really. I think we can do away with it entirely.

kcalvinalvin:
Rebased on top of main. Now the macros aren't moved to the tests but are just outright removed.

Davidson-Souza:
ack a075bcce

PR #15, opened 2022-10-20
Bring down MSRV to rust 1.41.0

kcalvinalvin:
These commits that are required to bring rustreexo into the rust bitcoin organization as MSRV of 1.41.0 is required for rust-bitcoin.

PR #16, opened 2022-10-27
Update deletion for caching

Davidson-Souza:
~~This PR depends on #13 and #15; hence it's marked as draft.~~

This finalizes the implementation of required support data for proof update, now we also return which nodes have been updated during deletion.
Required for #12.

### DIFF
--- a/test_values/cache_tests.json
+++ b/test_values/cache_tests.json
@@ -17,7 +17,9 @@
             "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
             "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
         ],
-        "to_destroy": []
+        "to_destroy": [],
+        "new_del_pos": [],
+        "new_del_hashes": []
     },
     {
         "initial_stump": "Add six leaves [0, 1, 2, 3, 4, 5]",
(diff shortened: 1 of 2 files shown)


test_values/cache_tests.json
```diff
@@ -137,6 +145,191 @@
...
+        ],
+        "leaves": 6,
+        "additional_preimages": [0, 1, 2, 3, 4],
+        "del_hashes": [
+            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
+            "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
+            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5"
+        ],
+        "proof_hashes": [
```
kcalvinalvin, line 166: The verification for this proof will pass since it needs no proof hashes to verify `[0, 1, 2, 3]` in a forest with 6 leaves. I'm not sure if we should treat these as a failure but since putting in any hashes here will pass the tests.

But we should really do it the correct way and have nothing here as the proofs.

kcalvinalvin (commented):
ACK 812a3f9af6081cce057be2964a09a3ff78e42f3c

Tested the new test cases. While they all pass the first one shouldn't have any `proof_hashes`. Everything else looks good.

test_values/cache_tests.json
Davidson-Souza: I'm not sure where this hashes came from. I removed them.

kcalvinalvin:
ACK 3ac22ed6d4f95decf52821ad5bde0861fe709bf3

PR #17, opened 2022-11-04
Implementation of Proof::Update

Davidson-Souza:
This PR finishes the required steps for #12. Now it's possible to update a proof independently given a set of addition and deletions from a block.

### DIFF
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -1,4 +1,5 @@
 use super::stump::UpdateData;
+use super::util::get_proof_positions;
 use crate::accumulator::{stump::Stump, types, util};
 use bitcoin_hashes::{sha256, Hash};
 
@@ -249,15 +250,154 @@ impl Proof {
     pub fn update(
         self,
         cached_hashes: Vec<sha256::Hash>,
+        add_hashes: Vec<sha256::Hash>,
         block_targets: Vec<u64>,
+        remembers: Vec<u64>,
         update_data: UpdateData,
     ) -> Result<(Proof, Vec<sha256::Hash>), String> {
-        self.update_proof_remove(
+        let (proof_after_deletion, cached_hashes) = self.update_proof_remove(
             block_targets,
             cached_hashes.clone(),
             update_data.new_del,
             update_data.prev_num_leaves,
-        )
+        )?;
+
+        let data_after_addition = proof_after_deletion.update_proof_add(
+            add_hashes,
+            cached_hashes.clone(),
+            remembers,
+            update_data.new_add,
+            update_data.prev_num_leaves,
+            update_data.to_destroy,
+        )?;
+
+        Ok(data_after_addition)
+    }
+    fn update_proof_add(
+        self,
+        adds: Vec<sha256::Hash>,
+        cached_del_hashes: Vec<sha256::Hash>,
+        remembers: Vec<u64>,
+        new_nodes: Vec<(u64, sha256::Hash)>,
+        before_num_leaves: u64,
+        to_destroy: Vec<u64>,
+    ) -> Result<(Proof, Vec<sha256::Hash>), String> {
+        // Combine the hashes with the targets.
+        let orig_targets_with_hash: Vec<(u64, sha256::Hash)> = self
+            .targets
+            .to_owned()
+            .into_iter()
+            .zip(cached_del_hashes.into_iter())
+            .collect();
+
+        // Attach positions to the proof.
+        let proof_pos = get_proof_positions(
+            &self.targets,
+            before_num_leaves,
+            util::tree_rows(before_num_leaves),
+        );
+        let proof_with_pos = proof_pos
+            .clone()
+            .into_iter()
+            .zip(self.hashes.clone())
+            .collect();
+
+        // Remap the positions if we moved up a after the addition row.
+        let targets_after_remap = Proof::maybe_remap(
+            before_num_leaves,
+            adds.len() as u64,
+            orig_targets_with_hash.clone(),
+        );
+        let mut final_targets = targets_after_remap.clone();
+        let mut new_nodes_iter = new_nodes.iter();
+        let mut proof_with_pos =
+            Proof::maybe_remap(before_num_leaves, adds.len() as u64, proof_with_pos);
+
+        let num_leaves = before_num_leaves + (adds.len() as u64);
+        // Move up positions that need to be moved up due to the empty roots
+        // being written over.
+        for node in to_destroy {
+            final_targets =
+                Proof::calc_next_positions(&vec![node], &final_targets, num_leaves, true)?;
+            proof_with_pos =
+                Proof::calc_next_positions(&vec![node], &proof_with_pos, num_leaves, true)?;
+        }
+
+        // remembers is an index telling what newly created UTXO should be cached
+        for remember in remembers {
+            let remember_target: Option<&sha256::Hash> = adds.get(remember as usize);
+            if let Some(remember_target) = remember_target {
+                let node = new_nodes_iter.find(|(_, hash)| *hash == *remember_target);
+                if let Some((pos, hash)) = node {
+                    final_targets.push((*pos, *hash));
+                }
+            }
+        }
+
+        final_targets.sort();
+
+        let (new_target_pos, target_hashes) = final_targets.clone().into_iter().unzip();
+        // Grab all the new nodes after this add.
+        let mut needed_proof_positions =
+            util::get_proof_positions(&new_target_pos, num_leaves, util::tree_rows(num_leaves));
+        needed_proof_positions.sort();
+
+        // We'll use all elements from the old proof, as addition only creates new nodes
+        // in our proof (except for root destruction). But before using it, we have to
+        // compute the new positions, as adding new elements may move existing elements a few
+        // rows up.
+        let mut new_proof = proof_with_pos;
+        // Iterates over the needed positions and grab them from new_nodes
+        // All proof elements must come from the old proof or new_nodes. Old proof elements
+        // are already in new_proof. Some every missing element must be in new_nodes.
+        for pos in needed_proof_positions {
+            if let Some((_, hash)) = new_nodes.iter().find(|(node_pos, _)| pos == *node_pos) {
+                new_proof.push((pos, *hash));
+            } else {
+                // This node must be in either new_nodes or in the old proof, otherwise we can't
+                // update our proof
+                if let None = new_proof.iter().find(|(proof_pos, _)| *proof_pos == pos) {
+                    return Err(format!("Missing position {}", pos));
+                }
+            }
+        }
+        new_proof.sort();
+
+        let (_, hashes): (Vec<u64>, Vec<sha256::Hash>) = new_proof.into_iter().unzip();
+        Ok((
+            Proof {
+                hashes,
+                targets: new_target_pos,
+            },
+            target_hashes,
+        ))
+    }
+    /// maybe_remap remaps the passed in hash and pos if the tree_rows increase after
+    /// adding the new nodes.
+    fn maybe_remap(
+        num_leaves: u64,
+        num_adds: u64,
+        positions: Vec<(u64, sha256::Hash)>,
+    ) -> Vec<(u64, sha256::Hash)> {
+        let new_forest_rows = util::tree_rows(num_leaves + num_adds);
+        let old_forest_rows = util::tree_rows(num_leaves);
+        let tree_rows = util::tree_rows(num_leaves);
+        let mut new_proofs = vec![];
+        if new_forest_rows > old_forest_rows {
+            for (pos, hash) in positions.iter() {
+                let row = util::detect_row(*pos, tree_rows);
+
+                let old_start_pos = util::start_position_at_row(row, old_forest_rows);
+                let new_start_pos = util::start_position_at_row(row, new_forest_rows);
+
+                let offset = pos - old_start_pos;
+                let new_pos = offset + new_start_pos;
+                new_proofs.push((new_pos, *hash));
+            }
+            return new_proofs;
+        }
+
+        positions
     }
 
     /// update_proof_remove modifies the cached proof with the deletions that happen in the block proof.
@@ -402,7 +542,7 @@ impl Proof {
 mod tests {
     use std::str::FromStr;
 
-    use bitcoin_hashes::{sha256, Hash, HashEngine};
+    use bitcoin_hashes::{hex::FromHex, sha256, Hash, HashEngine};
     use serde::Deserialize;
 
     use super::Proof;
@@ -416,7 +556,134 @@ mod tests {
         proofhashes: Vec<String>,
         expected: bool,
     }
+    /// This test checks whether our update proof works for different scenarios. We start
+    /// with a (valid) cached proof, then we receive `blocks`, like we would in normal Bitcoin
+    /// but for this test, block is just random data. For each block we update our Stump and
+    /// our proof as well, after that, our proof **must** still be valid for the latest Stump.
+    ///
+    /// Fix-me: Using derive for deserialize, when also using sha256::Hash leads to an odd
+    /// error that can't be easily fixed. Even bumping version doesn't appear to help.
+    /// Deriving hashes directly reduces the amount of boilerplate code used, and makes everything
+    /// more clearer, hence, it's preferable.
+    #[test]
+    fn test_update_proof() {
+        #[derive(Debug, Deserialize)]
+        struct JsonProof {
+            targets: Vec<u64>,
+            hashes: Vec<String>,
+        }
+        #[derive(Debug, Deserialize)]
+        struct UpdatedData {
+            /// The newly created utxo to be added to our accumulator
+            adds: Vec<u64>,
+            /// The proof for all destroyed utxos
+            proof: JsonProof,
+            /// The hash of all destroyed utxos
+            del_hashes: Vec<String>,
+        }
+        #[derive(Debug, Deserialize)]
+        struct TestData {
+            /// Blocks contains new utxos and utxos that are being deleted
+            update: UpdatedData,
+            /// The proof we have for our wallet's utxos
+            cached_proof: JsonProof,
+            /// A initial set of roots, may be empty for starting with an empty stump
+            initial_roots: Vec<String>,
+            /// The number of leaves in the initial Stump
+            initial_leaves: u64,
+            /// The hash of all wallet's utxo
+            cached_hashes: Vec<String>,
+            /// The indexes of all the new utxos to cache
+            remembers: Vec<u64>,
+            /// After we update our stump, which roots we expect?
+            expected_roots: Vec<String>,
+            /// After we update the proof, the proof's target should be this
+            expected_targets: Vec<u64>,
+            /// After we update the proof, the cached hashes should be this
+            expected_cached_hashes: Vec<String>,
+        }
+        let contents = std::fs::read_to_string("test_values/cached_proof_tests.json")
+            .expect("Something went wrong reading the file");
 
+        let values: Vec<TestData> =
+            serde_json::from_str(contents.as_str()).expect("JSON deserialization error");
+        for case_values in values {
+            let proof_hashes = case_values
+                .cached_proof
+                .hashes
+                .iter()
+                .map(|val| sha256::Hash::from_str(val).unwrap())
+                .collect();
+            let cached_hashes: Vec<_> = case_values
+                .cached_hashes
+                .iter()
+                .map(|val| sha256::Hash::from_str(val).unwrap())
+                .collect();
+
+            let cached_proof = Proof::new(case_values.cached_proof.targets, proof_hashes);
+            let roots = case_values
+                .initial_roots
+                .into_iter()
+                .map(|hash| sha256::Hash::from_hex(&hash).unwrap())
+                .collect();
+
+            let stump = Stump {
+                leafs: case_values.initial_leaves,
+                roots,
+            };
+
+            let utxos = case_values
+                .update
+                .adds
+                .iter()
+                .map(|preimage| hash_from_u8(*preimage as u8))
+                .collect();
+            let del_hashes = case_values
+                .update
+                .del_hashes
+                .iter()
+                .map(|hash| sha256::Hash::from_hex(hash).unwrap())
+                .collect();
+            let block_proof_hashes: Vec<_> = case_values
+                .update
+                .proof
+                .hashes
+                .iter()
+                .map(|hash| sha256::Hash::from_hex(hash).unwrap())
+                .collect();
+
+            let block_proof =
+                Proof::new(case_values.update.proof.targets.clone(), block_proof_hashes);
+            let (stump, updated) = stump.modify(&utxos, &del_hashes, &block_proof).unwrap();
+            let (cached_proof, cached_hashes) = cached_proof
+                .update(
+                    cached_hashes.clone(),
+                    utxos,
+                    case_values.update.proof.targets,
+                    case_values.remembers.clone(),
+                    updated.clone(),
+                )
+                .unwrap();
+
+            let res = cached_proof.verify(&cached_hashes, &stump);
+
+            let expected_roots: Vec<_> = case_values
+                .expected_roots
+                .iter()
+                .map(|hash| sha256::Hash::from_hex(hash).unwrap())
+                .collect();
+
+            let expected_cached_hashes: Vec<_> = case_values
+                .expected_cached_hashes
+                .iter()
+                .map(|hash| sha256::Hash::from_hex(hash).unwrap())
+                .collect();
+            assert_eq!(res, Ok(true));
+            assert_eq!(cached_proof.targets, case_values.expected_targets);
+            assert_eq!(stump.roots, expected_roots);
+            assert_eq!(cached_hashes, expected_cached_hashes);
+        }
+    }
     #[test]
     fn test_calc_next_positions() {
         use super::Proof;
--- a/test_values/cached_proof_tests.json
+++ b/test_values/cached_proof_tests.json
@@ -1,0 +1,379 @@
+[
+    {
+        "update": {
+            "adds": [20, 21, 22],
+            "proof": {
+                "targets": [0, 1],
+                "hashes": [
+                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
+                    "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7"
+                ]
+            },
+            "del_hashes": [
+                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+                "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a"
+            ]
+        },
+        "remembers": [],
+        "cached_proof": {
+            "targets": [4, 6, 7],
+            "hashes": [
+                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+            ]
+        },
+        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
+        "initial_roots": [
+            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
+        ],
+        "initial_leaves": 8,
+        "cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ],
+        "expected_roots": [
+            "97491b30a42410dc3267d17933cf5e1b55cfb92ebab2dcf1bcd098032dacee95",
+            "0d451c2d9366705a3557fd038c25a40a348cc0effa2dd00dfaaba65749cd0915",
+            "7cb7c4547cf2653590d7a9ace60cc623d25148adfbc88a89aeb0ef88da7839ba"
+        ],
+        "expected_targets": [4, 6, 7],
+        "expected_cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ]
+    },
+    {
+        "update": {
+            "adds": [8],
+            "proof": {
+                "targets": [0, 1, 4],
+                "hashes": [
+                    "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
+                    "34028bbc87000c39476cdc60cf80ca32d579b3a0e2d3f80e0ad8c3739a01aa91"
+                ]
+            },
+            "del_hashes": [
+                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+                "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
+                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71"
+            ]
+        },
+        "remembers": [0],
+        "cached_proof": {
+            "targets": [4, 6, 7],
+            "hashes": [
+                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+            ]
+        },
+        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
+        "initial_roots": [
+            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
+        ],
+        "initial_leaves": 8,
+        "cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ],
+        "expected_roots": [
+            "7f6df6757821144278f5d535261ab4ffc55e2618f7bc62f1a19e7d289061b08b",
+            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a"
+        ],
+        "expected_targets": [6, 7, 8],
+        "expected_cached_hashes": [
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
+            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a"
+        ]
+    },
+    {
+        "update": {
+            "adds": [8],
+            "proof": {
+                "targets": [4],
+                "hashes": [
+                    "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                    "34028bbc87000c39476cdc60cf80ca32d579b3a0e2d3f80e0ad8c3739a01aa91",
+                    "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+                ]
+            },
+            "del_hashes": [
+                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71"
+            ]
+        },
+        "remembers": [0],
+        "cached_proof": {
+            "targets": [4, 5, 7],
+            "hashes": [
+                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+                "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+            ]
+        },
+        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
+        "initial_roots": [
+            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
+        ],
+        "initial_leaves": 8,
+        "cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ],
+        "expected_roots": [
+            "f1e8f77fb6c03d3e62acc1824e6671da21bb643f362a9af5171167bfdfbbb9cd",
+            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a"
+        ],
+        "expected_targets": [7, 8, 18],
+        "expected_cached_hashes": [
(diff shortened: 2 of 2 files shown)


test_values/cached_proof_tests.json
```diff
@@ -0,0 +1,41 @@
...
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ],
+        "expected_roots": [
+            "97491b30a42410dc3267d17933cf5e1b55cfb92ebab2dcf1bcd098032dacee95",
+            "0d451c2d9366705a3557fd038c25a40a348cc0effa2dd00dfaaba65749cd0915",
+            "7cb7c4547cf2653590d7a9ace60cc623d25148adfbc88a89aeb0ef88da7839ba"
+        ],
+        "expected_targets": [4, 6, 7]
+    }
+]
```
kcalvinalvin, line 41: The below test case should pass:

```
    {
        "update": {
            "adds": [8],
            "proof": {
                "targets": [0, 1, 4],
                "hashes": [
                    "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                    "34028bbc87000c39476cdc60cf80ca32d579b3a0e2d3f80e0ad8c3739a01aa91"
                ]
            },
            "del_hashes": [
                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
                "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71"
            ]
        },
        "remembers": [0],
        "cached_proof": {
            "targets": [4, 6, 7],
            "hashes": [
                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
                "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
            ]
        },
        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
        "initial_roots": [
            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
        ],
        "initial_leaves": 8,
        "cached_hashes": [
            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
        ],
        "expected_roots": [
            "7f6df6757821144278f5d535261ab4ffc55e2618f7bc62f1a19e7d289061b08b",
            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d"
        ],
        "expected_targets": [6, 7, 8],
        "expected_cached_hashes": [
                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
                "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d"
        ]
    }
```

diff of the `cached_proof_tests.json`:
```diff
diff --git a/test_values/cached_proof_tests.json b/test_values/cached_proof_tests.json
index 6b27f11..bf4e1c2 100644
--- a/test_values/cached_proof_tests.json
+++ b/test_values/cached_proof_tests.json
@@ -14,6 +14,7 @@
                 "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a"
             ]
         },
+        "remembers": [],
         "cached_proof": {
             "targets": [4, 6, 7],
             "hashes": [
@@ -36,6 +37,57 @@
             "0d451c2d9366705a3557fd038c25a40a348cc0effa2dd00dfaaba65749cd0915",
             "7cb7c4547cf2653590d7a9ace60cc623d25148adfbc88a89aeb0ef88da7839ba"
         ],
-        "expected_targets": [4, 6, 7]
+        "expected_targets": [4, 6, 7],
+        "expected_cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ]
+    },
+    {
+        "update": {
+            "adds": [8],
+            "proof": {
+                "targets": [0, 1, 4],
+                "hashes": [
+                    "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
+                    "34028bbc87000c39476cdc60cf80ca32d579b3a0e2d3f80e0ad8c3739a01aa91"
+                ]
+            },
+            "del_hashes": [
+                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
+                "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
+                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71"
+            ]
+        },
+        "remembers": [0],
+        "cached_proof": {
+            "targets": [4, 6, 7],
+            "hashes": [
+                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
+                "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
+            ]
+        },
+        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
+        "initial_roots": [
+            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
+        ],
+        "initial_leaves": 8,
+        "cached_hashes": [
+            "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
+            "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
+        ],
+        "expected_roots": [
+            "7f6df6757821144278f5d535261ab4ffc55e2618f7bc62f1a19e7d289061b08b",
+            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d"
+        ],
+        "expected_targets": [6, 7, 8],
+        "expected_cached_hashes": [
+                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
+                "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
+                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d"
+        ]
     }  
-]      
\ No newline at end of file
+]
```

diff of the `proof.rs` file to support the test case:
```diff
diff --git a/src/accumulator/proof.rs b/src/accumulator/proof.rs
index 8bac7ba..8e5e70e 100644
--- a/src/accumulator/proof.rs
+++ b/src/accumulator/proof.rs
@@ -568,10 +568,14 @@ mod tests {
             initial_leaves: u64,
             /// The hash of all wallet's utxo
             cached_hashes: Vec<String>,
+            /// The indexes of all the new utxos to cache
+            remembers: Vec<u64>,
             /// After we update our stump, which roots we expect?
             expected_roots: Vec<String>,
             /// After we update the proof, the proof's target should be this
             expected_targets: Vec<u64>,
+            /// After we update the proof, the cached hashes should be this
+            expected_cached_hashes: Vec<String>,
         }
         let contents = std::fs::read_to_string("test_values/cached_proof_tests.json")
             .expect("Something went wrong reading the file");
@@ -632,7 +636,7 @@ mod tests {
                     cached_hashes.clone(),
                     utxos,
                     case_values.update.proof.targets,
-                    vec![],
+                    case_values.remembers,
                     updated.clone(),
                 )
                 .unwrap();
@@ -645,9 +649,16 @@ mod tests {
                 .map(|hash| sha256::Hash::from_hex(hash).unwrap())
                 .collect();
 
+            let expected_cached_hashes: Vec<_> = case_values
+                .expected_cached_hashes
+                .iter()
+                .map(|hash| sha256::Hash::from_hex(hash).unwrap())
+                .collect();
+
             assert_eq!(res, Ok(true));
             assert_eq!(cached_proof.targets, case_values.expected_targets);
             assert_eq!(stump.roots, expected_roots);
+            assert_eq!(cached_hashes, expected_cached_hashes);
         }
     }
     #[test]
```

kcalvinalvin (commented):
Overall I feel like this PR needs more work because of the cached_hashes not being updated.

If you want, you could separate the functions:

`start_position_at_row`
`maybe_remap`
`calc_next_positions`

since those look ok to be merged.

test_values/cached_proof_tests.json
Davidson-Souza: Ok, I think I get this working. I've misinterpreted what `remembers` is, it doesn't says what nodes should be remembered, but what UTXO from the set of new additions in this block should be remembered. I'm also able to return `cached_hashes` from update.   
But I have one question about this test case. You add a leaf with value `8` and uses `6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d` as its hash (the lower root and cached hashes has this value). But `6e34` is the hash of `0`. If I'm not mistaken, `8` hashes to `beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a`. Did I miss something on this `add` value?

kcalvinalvin: Ok yeah you're right. 8 hashes to `beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a`.

There was a bug in my go code to generate this. 8 should be `beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a`.

Davidson-Souza:
Force-pushed ba88aa1, addressing requests. Now we return an updated `cached_hashes` vector.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,270 @@ impl Proof {
...
+                let row = util::detect_row(*pos, tree_rows);
+
+                let old_start_pos = util::start_position_at_row(row, old_forest_rows);
+                let new_start_pos = util::start_position_at_row(row, new_forest_rows);
+
+                let offset = pos - old_start_pos;
+                let new_pos = offset + new_start_pos;
+                new_proofs.push((new_pos, hash));
+            }
+            return positions;
```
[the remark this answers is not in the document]
Davidson-Souza: Fixed!

src/accumulator/proof.rs
```diff
@@ -241,6 +243,270 @@ impl Proof {
...
+                .find(|(updated_pos, _)| missing == *updated_pos)
+            {
+                if *hash != sha256::Hash::default() {
+                    new_proof.push((missing, *hash));
+                }
+            }
+        }
+        let (_, hashes): (Vec<u64>, Vec<sha256::Hash>) = new_proof.into_iter().unzip();
+        let targets =
+            Proof::calc_next_positions(&block_targets, &targets_with_hash, num_leaves, false)?;
```
kcalvinalvin, line 466: `calc_next_positions()` should also be called on the proofs since they'll also change positions.

Here's a test case that makes the current code fail.

```
    {
        "update": {
            "adds": [],
            "proof": {
                "targets": [4, 5, 6, 7],
                "hashes": [
                    "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386"
                ]
            },
            "del_hashes": [
                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
                "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
            ]
        },
        "remembers": [],
        "cached_proof": {
            "targets": [1, 8, 12],
            "hashes": [
                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
                "2b4c342f5433ebe591a1da77e013d1b72475562d48578dca8b84bac6651c3cb9",
                "9d1e0e2d9459d06523ad13e28a4093c2316baafe7aec5b25f30eba2e113599c4",
                "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                "c413035120e8c9b0ca3e40c93d06fe60a0d056866138300bb1f1dd172b4923c3",
                "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7"
            ]
        },
        "initial_stump": "Add 14 leaves [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]",
        "initial_roots": [
            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42",
            "9c053db406c1a077112189469a3aca0573d3481bef09fa3d2eda3304d7d44be8",
            "55d0a0ef8f5c25a9da266b36c0c5f4b31008ece82df2512c8966bddcc27a66a0"
        ],
        "initial_leaves": 14,
        "cached_hashes": [
            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a",
            "ef6cbd2161eaea7943ce8693b9824d23d1793ffb1c0fca05b600d3899b44c977"
        ],
        "expected_roots": [
            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386",
            "9c053db406c1a077112189469a3aca0573d3481bef09fa3d2eda3304d7d44be8",
            "55d0a0ef8f5c25a9da266b36c0c5f4b31008ece82df2512c8966bddcc27a66a0"
        ],
        "expected_targets": [8, 12, 17],
        "expected_cached_hashes": [
            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a",
            "ef6cbd2161eaea7943ce8693b9824d23d1793ffb1c0fca05b600d3899b44c977",
            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a"
        ]
    }
```

Davidson-Souza: Right, if a proof node gets moved up, it might make the proof in the wrong order. Pushing a fix.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,286 @@ impl Proof {
...
+                }
+            }
+        }
+
+        // We need to remap all proof hashes and sort then, otherwise our hash will be wrong.
+        // This happens because deletion moves nodes upwards, some of this nodes may be a proof
+        // element. If so we move it to its new position. After this, we might have duplicates, and
+        // the vector is probably unsorted, so we dedup and sort it again.
+        let (pos, hashes): (Vec<u64>, Vec<sha256::Hash>) =
+            Proof::calc_next_positions(&block_targets, &new_proof, num_leaves, false)?;
```
kcalvinalvin, line 470: There shouldn't be any duplicates resulting from `calc_next_positions`.

I found that the `missing_positions` variable is wrong, leading to duplicates being added. No need to dedupe here. Maybe need to sort? I'm not sure if the code here guarantees sorted-ness.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,286 @@ impl Proof {
...
+                    new_proof.push((*pos, *updated_hash));
+                }
+            } else {
+                if hash != sha256::Hash::default() {
+                    new_proof.push((*pos, hash));
+                }
+            }
+        }
+
+        let missing_positions = needed_pos
```
kcalvinalvin, line 449: `missing_positions` are the resulting vector of `u64`s from `get_proof_positions` minus the following:

1. Deleted targets in the current block. These would be `block_targets`.
2. All the current positions we're aware of. These would be included in `new_proof`. (I realized the go impl basically uses the result from proof_positions. This still would work since `new_proof` is a subset of `proof_positions` but we should be using `new_proof`).

We're not getting rid of the `block_targets` and that's why we're getting duplicates. Should add a filter for `block_targets` here.

Davidson-Souza: I think the problem was the filter closure. I should be: 
```rust
            .filter(|pos| !proof_positions.contains(pos));
```
But I missed the negation of `proof_positions.contains`, this brings elements that already are in the proof to `missing`, which is wrong.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,286 @@ impl Proof {
...
+        let old_proof: Vec<_> = proof_positions
+            .iter()
+            .zip(self.hashes.clone().into_iter())
+            .collect();
+
+        let mut new_proof = vec![];
+        // Grab all the positions of the needed proof hashes.
+        let needed_pos = util::get_proof_positions(&targets, num_leaves, total_rows);
+
+        let updated: Vec<_> = updated
```
kcalvinalvin, line 427: We should be creating a `extra_positions` variable instead of shadowing `updated` since we'll need it for adding new proofs.

Take for example this tree:
```
14
|---------------\
12              13
|-------\       |-------\
08      09      10      11
|---\   |---\   |---\   |---\
00  01  02  03  04  05  06  07
```

If we cache `[00, 07]`, the hashes of `12` and `13` are left out as they're able to be calculated from hashing either `(00 | 01) | 09` or `10 | (06 | 07)`. But if we delete `0`, then `12` isn't able to be calculated.

`updated` provides the new hash for `12` after the deletion of `0` and we use this data to add in the missing `12`. But the code here filters out `12`, so this ends up creating an invalid proof.

Here's the scenario I described above as a test case:

```
    {
        "update": {
            "adds": [],
            "proof": {
                "targets": [0],
                "hashes": [
                    "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
                    "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                    "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7"
                ]
            },
            "del_hashes": [
                "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d"
            ]
        },
        "remembers": [],
        "cached_proof": {
            "targets": [0, 7],
            "hashes": [
                "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6",
                "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                "9eec588c41d87b16b0ee226cb38da3864f9537632321d8be855a73d5616dcc73"
            ]
        },
        "initial_stump": "Add 8 leaves [0, 1, 2, 3, 4, 5, 6, 7]",
        "initial_roots": [
            "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42"
        ],
        "initial_leaves": 8,
        "cached_hashes": [
            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
        ],
        "expected_roots": [
            "726fdd3b432cc59e68487d126e70f0db74a236267f8daeae30b31839a4e7ebed"
        ],
        "expected_targets": [7],
        "expected_cached_hashes": [
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
        ]
    }
```

kcalvinalvin (commented):
I'd like to have comments above each block of code that explains what the code is doing for `update_proof_add` and `update_proof_remove`. It'd be helpful for documentation/reading the code.

src/accumulator/proof.rs
Davidson-Souza: I changed how we deal with old proof, now we use `updated` without filtering, but only asks for positions that are required for the new proof. Because using `updated` as passed in would create nodes that are no longer needed. In this case, `01` and `09`. The filtered version is not used anymore.

kcalvinalvin (commented):
While I was coming up with the test case to make not sorting targets fail, I also came across another case that should be passing but is failing in the current code. I've not looked at why this fails but it should pass.

```
    {
        "update": {
            "adds": [],
            "proof": {
                "targets": [3, 6],
                "hashes": [
                    "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
                    "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
                    "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
                    "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db"
                ]
            },
            "del_hashes": [
                "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5",
                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6"
            ]
        },
        "remembers": [],
        "cached_proof": {
            "targets": [2, 3, 7, 16, 18, 26],
            "hashes": [
                "67586e98fad27da0b9968bc039a1ef34c939b9b8e523a8bef89d478608c5ecf6"
            ]
        },
        "initial_stump": "Add 12 leaves [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11], then delete [1, 4, 8, 9, 10]",
        "initial_roots": [
            "5093a5ae2bceeef66d9af3f4f1ed71acffe4932cd02afca5d2dc659dca5c418f",
            "e7cf46a078fed4fafd0b5e3aff144802b853f8ae459a4f0c14add3314b7cc3a6"
        ],
        "initial_leaves": 12,
        "cached_hashes": [
            "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5",
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
            "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
            "e7cf46a078fed4fafd0b5e3aff144802b853f8ae459a4f0c14add3314b7cc3a6"
        ],
        "expected_roots": [
            "2bfc2e9199263e7b2ef762afeb2fede84eb4fac37f3f28fd22d5761c8390b875",
            "e7cf46a078fed4fafd0b5e3aff144802b853f8ae459a4f0c14add3314b7cc3a6"
        ],
        "expected_targets": [16, 17, 18, 19, 26],
        "expected_cached_hashes": [
            "6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d",
            "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
            "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db",
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879",
            "e7cf46a078fed4fafd0b5e3aff144802b853f8ae459a4f0c14add3314b7cc3a6"
        ]
    }
```

Here's the go code I used to generate the above test case: https://go.dev/play/p/dSFevPamlW3

src/accumulator/proof.rs
```diff
@@ -241,6 +243,284 @@ impl Proof {
...
+            Proof::calc_next_positions(&block_targets, &new_proof, num_leaves, false)?;
+
+        let mut proof_elements: Vec<_> = pos.into_iter().zip(hashes.into_iter()).collect();
+        proof_elements.sort();
+
+        let (_, hashes): (Vec<u64>, Vec<sha256::Hash>) = proof_elements.into_iter().unzip();
+        let (targets, _) =
+            Proof::calc_next_positions(&block_targets, &targets_with_hash, num_leaves, false)?;
+
+        Ok((Proof { hashes, targets }, target_hashes))
```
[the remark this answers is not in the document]
Davidson-Souza: I could find where the problem with the first is. `get_next_pos` didn't return roots, but 26 is as root. I did manage to sort all targets and target hashes, but this last test case looks a little weird, the cached proof hashes and cached hashes are really equal?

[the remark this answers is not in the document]
kcalvinalvin: > I could find where the problem with the first is. `get_next_pos` didn't return roots, but 26 is as root. I did manage to sort all targets and target hashes, but this last test case looks a little weird, the cached proof hashes and cached hashes are really equal?

Oh yeah you're right. I copied the wrong thing for the proof hashes. Edited the original comment to fix that.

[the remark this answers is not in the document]
Davidson-Souza: Ok, now it passes in both cases.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,291 @@ impl Proof {
...
+            final_targets.clone().into_iter().unzip();
+        let mut new_proof_pos = proof_pos;
+
+        for node in to_destroy {
+            (new_target_pos, _) =
+                Proof::calc_next_positions(&vec![node], &final_targets, num_leaves, true)?;
+            (new_proof_pos, _) =
+                Proof::calc_next_positions(&vec![node], &proof_with_pos, num_leaves, true)?;
+        }
+        // Add in proof_hashes to the new_nodes as some needed hashes
```
kcalvinalvin, line 340: Where are the proof_hashes being added in? I don't see it in the code

src/accumulator/proof.rs
```diff
@@ -241,6 +243,291 @@ impl Proof {
...
+            if let Some(remember_target) = remember_target {
+                let node = new_nodes_iter.find(|(_, hash)| *hash == *remember_target);
+                if let Some((pos, hash)) = node {
+                    final_targets.push((*pos, *hash));
+                }
+            }
+        }
+
+        final_targets.sort();
+        // Move up positions that need to be moved up due to the empty roots
```
kcalvinalvin, line 327: This isn't where the positions move up though. Doesn't look like the comment is describing what the code is doing

src/accumulator/proof.rs
```diff
@@ -241,6 +243,291 @@ impl Proof {
...
+        // Grab all the new nodes after this add.
+        let needed_proof_positions = util::get_proof_positions(
+            &new_target_pos,
+            leaves_after_add,
+            util::tree_rows(leaves_after_add),
+        );
+        let mut new_proof: Vec<_> = new_proof_pos.into_iter().zip(self.hashes).collect();
+        for pos in needed_proof_positions {
+            for (new_node_pos, hash) in new_nodes.iter() {
+                if *new_node_pos == pos {
```
kcalvinalvin, line 352: There should be an error check here for when we don't find the needed position.

The vec of nodes should always contain all the needed positions

kcalvinalvin (changes requested):

src/accumulator/proof.rs
Davidson-Souza: Do you mean in the new proof?

src/accumulator/proof.rs
Davidson-Souza: When I add an else clause with an error, it breaks. I am not sure why.

src/accumulator/proof.rs
kcalvinalvin: I mean the old proof hashes (that we've updated with new positions).

In the go impl I create a completely new proof object so you need to also grab the old proof hashes as well. Looks like you're doing that part in a different way.

The comment isn't really describing what the code is doing so it should be removed.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,291 @@ impl Proof {
...
+        // Add in proof_hashes to the new_nodes as some needed hashes
+        // will be in the proof hashes.
+        let leaves_after_add = ((before_num_leaves as usize) + adds.len()) as u64;
+        // Grab all the new nodes after this add.
+        let needed_proof_positions = util::get_proof_positions(
+            &new_target_pos,
+            leaves_after_add,
+            util::tree_rows(leaves_after_add),
+        );
+        let mut new_proof: Vec<_> = new_proof_pos.into_iter().zip(self.hashes).collect();
```
kcalvinalvin, line 349: `new_proof_pos` is guaranteed to be sorted. Some of the positions may move around so it's not guaranteed to be in the same order as `self.hashes`.

In order for the hashes to be in the same order as the positions, we need to update the ordering of the hashes when we shuffle the positions around as well.

src/accumulator/proof.rs
kcalvinalvin: Could you post the code you used and the error that occurs when you made the change?

Davidson-Souza: Sure. This is the `for` loop in line 350.  
```rust
        for pos in needed_proof_positions {
            for (new_node_pos, hash) in new_nodes.iter() {
                if *new_node_pos == pos {
                    new_proof.push((pos, *hash));
                } else {
                    return Err(format!("Missing position {pos}"));
                }
            }
        }
```

src/accumulator/proof.rs
Davidson-Souza: I think it'll be in the same position because `calc_next_positions` doesn't move targets, the sorting happens below with both hashes and targets.

kcalvinalvin: Ah that's another bug then. `calc_next_positions` should result in positions being moved around.

I made this test case for `calc_next_positions`. It's currently failing on the latest commit.

```
    #[test]
    fn test_calc_next_positions() {
        use super::Proof;

        #[derive(Clone)]
        struct Test {
            name: &'static str,
            block_targets: Vec<u64>,
            old_positions: Vec<(u64, sha256::Hash)>,
            num_leaves: u64,
            num_adds: u64,
            append_roots: bool,
            expected: (Vec<u64>, Vec<sha256::Hash>),
        }

        let tests = vec![Test {
            name: "One empty root deleted",
            block_targets: vec![26],
            old_positions: vec![
                (
                    1,
                    bitcoin_hashes::sha256::Hash::from_str(
                        "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
                    )
                    .unwrap(),
                ),
                (
                    13,
                    bitcoin_hashes::sha256::Hash::from_str(
                        "9d1e0e2d9459d06523ad13e28a4093c2316baafe7aec5b25f30eba2e113599c4",
                    )
                    .unwrap(),
                ),
                (
                    17,
                    bitcoin_hashes::sha256::Hash::from_str(
                        "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                    )
                    .unwrap(),
                ),
                (
                    25,
                    bitcoin_hashes::sha256::Hash::from_str(
                        "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7",
                    )
                    .unwrap(),
                ),
            ],
            num_leaves: 14,
            num_adds: 14,
            append_roots: false,
            expected: (
                vec![1, 17, 21, 25],
                vec![
                    bitcoin_hashes::sha256::Hash::from_str(
                        "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
                    )
                    .unwrap(),
                    bitcoin_hashes::sha256::Hash::from_str(
                        "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
                    )
                    .unwrap(),
                    bitcoin_hashes::sha256::Hash::from_str(
                        "9d1e0e2d9459d06523ad13e28a4093c2316baafe7aec5b25f30eba2e113599c4",
                    )
                    .unwrap(),
                    bitcoin_hashes::sha256::Hash::from_str(
                        "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7",
                    )
                    .unwrap(),
                ],
            ),
        }];

        for test in tests {
            let res = Proof::calc_next_positions(
                &test.block_targets,
                &test.old_positions,
                test.num_leaves + test.num_adds,
                test.append_roots,
            )
            .unwrap();

            assert_eq!(res, test.expected, "testcase: \"{}\" fail", test.name);
        }
    }
```

src/accumulator/proof.rs
```diff
@@ -269,6 +556,246 @@ mod tests {
...
+            .unwrap();
+        let cached_target_hashes: Vec<_> = vec![0, 1, 7]
+            .into_iter()
+            .map(|preimage| hash_from_u8(preimage))
+            .collect();
+
+        let res = cached_proof.update_proof_add(
+            additional_hashes,
+            cached_target_hashes.clone(),
+            vec![10],
```
kcalvinalvin, line 730: I think you wanted to cached the hash of `10` so it should be `vec![0]`

src/accumulator/proof.rs
kcalvinalvin: Ah ok so this is related to https://github.com/mit-dci/rustreexo/pull/17/files#r1029246624

For positions: `[0, 1, 7, 10]`, you need `[6, 11, 17, 18, 20, 27]`. In that loop, it'll try to grab every single one of the newly created positions starting from `6` to `27`.

However, `6` isn't part of the newly created positions. It's part of the old proofs that existed before this block.

In the go version, I'm creating a whole new proof so what I did was I added the old proofs to the `new_nodes`. In yours, the `new_proof` already contains `6`, along with `17` and `18`.

So I guess you need to filter out what you already have from the `needed_proof_positions`.

`needed_proof_positions` = `[6, 11, 17, 18, 20, 27]` but since you already have `[6, 17, 18]` in `new_proof`, `needed_proof_positions` can just be `[11, 20, 27]`.

src/accumulator/proof.rs
Davidson-Souza: > Ah that's another bug then. calc_next_positions should result in positions being moved around.  

Ok, it does result in positions being moved around, but it doesn't return then sorted. I made this way because in some places I don't need hashes, only positions. To avoid unzips, `calc_next_positions` returns a tuple of `Vec<u64>` and `Vec<Hash>`. If I have to sort, then I need to hold a `Vec<(u64, Hash)>`, since I can't sort hashes without positions.  
The code is handling this by sorting if needed. Do you think it's worth changing `calc_next_positions` to produce a sorted `Vec<(u64, Hash)>`?

Davidson-Souza:
Force-push 4635eee:  
Adds an error if a proof node is missing in `update_proof_add`  
Removes duplicated test  
Add a few comments in the code

src/accumulator/proof.rs
kcalvinalvin: > Do you think it's worth changing `calc_next_positions` to produce a sorted `Vec<(u64, Hash)>`?

When I redid the algorithm for how the proof updates were done, I tried to avoid sorting wherever possible but I couldn't think of how to avoid sorting during position updates. So we can't avoid sorting (for now, maybe there is an algorithm that you can do without sorting).

You could internally use `Vec<(u64, Hash)>` and just unzip before returning. If unzip is too expensive when benchmarked, we can look at ways of reducing the cost.

Davidson-Souza: Ok, 826f047 sorts inside `calc_next_position`, I added your test, and it's passing.

kcalvinalvin (commented):
`update_proof_remove` looks good (at least I can't find any errors). `update_proof_add` needs a bit more touching up.

src/accumulator/proof.rs
```diff
@@ -241,6 +243,293 @@ impl Proof {
...
+            before_num_leaves,
+            adds.len() as u64,
+            orig_targets_with_hash.clone(),
+        );
+        let mut final_targets = targets_after_remap.clone();
+        let mut new_nodes_iter = new_nodes.iter();
+        let proof_with_pos =
+            Proof::maybe_remap(before_num_leaves, adds.len() as u64, proof_with_pos);
+        // remembers is an index telling what newly created UTXO should be cached
+        for remember in remembers {
```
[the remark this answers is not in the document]
Davidson-Souza: Those hashes look a little weird, `Add 6 leaves [0, 1, 2, 3, 4, 5], then delete [4, 5] and add [6, 7, 8]` and `delHashes` is 
```
040000ff00000000000000000000000000000000000000000000000000000000
050000ff00000000000000000000000000000000000000000000000000000000
```
ff000004 and ff000005 in little-endian, and 31 bytes with 0's. I'm almost certain this is not their hashes.  

If you meant to use this hash as leaf hash anyway without knowing the preimage, this won't work with this testing framework we made because it hashes the preimages (usually numbers) and use this as leaf hash.

kcalvinalvin:
Hey totally forgot about this PR.

Looks like I left some comments about update_proof_add but saw update_proof_delete was ok.

I could re-review this but do you wanna refactor out the update_proof_delete first into a different PR and get that merged?

Davidson-Souza:
> Hey totally forgot about this PR.   

I was about to ask you about this :laughing:   

> I could re-review this but do you wanna refactor out the update_proof_delete first into a different PR and get that merged?  

Yes, I can separate it into two prs.

kcalvinalvin:
Could you rebase this onto main? I'll take a look at the add then.

Davidson-Souza:
Rebased. IIRC the last thing missing to `add` is this https://github.com/mit-dci/rustreexo/pull/17#discussion_r1044226510

src/accumulator/proof.rs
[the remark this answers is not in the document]
kcalvinalvin: Ah the 0xff's are shoved in there in the go tests because go library interprets alls 0s as nothing. So instead of hashing it, the test just picks a byte and turns it into 0xff.

I'll change the test so it's better suited for this lib. But the:
```
040000ff00000000000000000000000000000000000000000000000000000000
050000ff00000000000000000000000000000000000000000000000000000000
```
 is supposed to represent `4` and `5`.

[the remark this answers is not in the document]
kcalvinalvin: Could you add this into the tests? This is the same as above but it's hashing the ints.

```
    {
        "update": {
            "adds": [6, 7, 8],
            "proof": {
                "targets": [4, 5],
                "hashes": []
            },
            "del_hashes": [
                "e52d9c508c502347344d8c07ad91cbd6068afc75ff6292f062a09ca381c89e71",
                "e77b9a9ae9e30b0dbdb6f510a264ef9de781501d7b6b92ae89eb059c5ab743db"
            ]
        },
        "remembers": [1],
        "cached_proof": {
            "targets": [3],
            "hashes": [
                "dbc1b4c900ffe48d575b5da5c638040125f65db0fe3e24494b76ea986457d986",
                "02242b37d8e851f1e86f46790298c7097df06893d6226b7c1453c213e91717de"
            ]
        },
        "initial_stump": "Add 6 leaves [0, 1, 2, 3, 4, 5], then delete [4, 5] and add [6, 7, 8]",
        "initial_roots": [
            "df46b17be5f66f0750a4b3efa26d4679db170a72d41eb56c3e4ff75a58c65386",
            "9eec588c41d87b16b0ee226cb38da3864f9537632321d8be855a73d5616dcc73"
        ],
        "initial_leaves": 6,
        "cached_hashes": [
            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5"
        ],
        "expected_roots": [
            "0f1ff6dc2fc7cf8e27236d51de9af7d32adf10a94eef498e167d5e7cf98b0112",
            "beead77994cf573341ec17b58bbf7eb34d2711c993c1d976b128b3188dc1829a"
        ],
        "expected_targets": [3, 19],
        "expected_cached_hashes": [
            "084fed08b978af4d7d196a7446a86b58009e636b611db16211b65a9aadff29c5",
            "ca358758f6d27e6cf45272937977a748fd88391db679ceda7dc7bf1f005ee879"
        ]
    }
```

src/accumulator/proof.rs
```diff
@@ -249,15 +250,154 @@ impl Proof {
...
+            before_num_leaves,
+            util::tree_rows(before_num_leaves),
+        );
+        let proof_with_pos = proof_pos
+            .clone()
+            .into_iter()
+            .zip(self.hashes.clone())
+            .collect();
+
+        // Remap the positions if we moved up a after the addition row.
```
kcalvinalvin, line 305: nit: Could you change the comment to:
`Remap the positions if we moved up a row after the addition.`

Just making things grammatically correct

src/accumulator/proof.rs
```diff
@@ -249,15 +250,154 @@ impl Proof {
...
+        // Move up positions that need to be moved up due to the empty roots
+        // being written over.
+        for node in to_destroy {
+            final_targets =
+                Proof::calc_next_positions(&vec![node], &final_targets, num_leaves, true)?;
+            proof_with_pos =
+                Proof::calc_next_positions(&vec![node], &proof_with_pos, num_leaves, true)?;
+        }
+
+        // remembers is an index telling what newly created UTXO should be cached
```
kcalvinalvin, line 326: nit: Could you change `UTXO` to `UTXOs`

kcalvinalvin (commented):
Things look good. Just these nits and the test case I mentioned (https://github.com/mit-dci/rustreexo/pull/17#discussion_r1083725464) being addressed should make this good to merge.

kcalvinalvin:
Hmm ok I could address the nits in a later PR. I'll merge it so it's not just sitting around forever.

Davidson-Souza:
I forgot to push that :man_facepalming: ? Sorry about that, I changed it locally and didn't push :')

PR #19, opened 2022-11-16
Remove unused transform file

kcalvinalvin:
transform function was used for the old utreexo accumulator algorithm. Since we've found a faster and more proof efficient algorithm, we can get rid of the entire transform file.


kcalvinalvin:
Paging @Davidson-Souza for review

Davidson-Souza:
ack [1e69e0f](https://github.com/mit-dci/rustreexo/pull/19/commits/1e69e0f01e8572d20bb5fed6b9d2ef02921fda7a). This code is not used in the codebase, so removing it makes sense.

PR #20, opened 2022-11-22
Sort before using targets in get_proof_positions

kcalvinalvin:
get_proof_positions doesn't enforce that the passed in targets should be sorted. However, the actual algorithm used does expect that the targets be sorted.

To ensure correctness, we need sort the targets that we've received before doing anything with it.

### DIFF
--- a/src/accumulator/util.rs
+++ b/src/accumulator/util.rs
@@ -32,7 +32,10 @@ pub fn calc_next_pos(position: u64, del_pos: u64, forest_rows: u8) -> Result<u64
     let pos_row = detect_row(position, forest_rows);
 
     if del_row < pos_row {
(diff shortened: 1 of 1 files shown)


src/accumulator/util.rs
```diff
@@ -391,6 +396,22 @@ mod tests {
 
     use bitcoin_hashes::sha256;
 
+    #[test]
```
Davidson-Souza, line 399: nit: Can you bring this test down in the file? There are some imports afterwards, I think it's clearer if imports come before any code.

kcalvinalvin: Addressed. I also moved up another import that was declared after a test code.

Davidson-Souza:
ack 7ca40d42

PR #21, opened 2023-01-11
Refactor: Optimize roots_to_destroy

Davidson-Souza:
**Chasing conceptual review on using lazy-static**
`roots_to_destroy` uses a non-zero hash as a marker, holding a root position that doesn't exist. We accomplish this by calling Hash::from_hex() and 0x01 as a 32-byte vector. However, from_hex involves expensive string parsing. After some testing[1], I found this function using more than 60% of cpu time while updating an accumulator.

This PR solves it by introducing a lazy global variable that contains this hash already, so we don't have to compute it over and over

[1] https://github.com/Davidson-Souza/dotfiles/blob/main/utreexo-wallet-flamegraph12212022.svg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "112c678d4050afce233f4f2852bb2eb519230b3cf12f33585275537d7e41578d"
 
 [[package]]
+name = "lazy_static"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
+
+[[package]]
 name = "libc"
 version = "0.2.126"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -97,6 +103,7 @@ name = "rustreexo"
 version = "0.1.0"
 dependencies = [
  "bitcoin_hashes",
+ "lazy_static",
  "serde",
  "serde_json",
  "sha2",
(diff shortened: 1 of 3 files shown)


kcalvinalvin:
> **Chasing conceptual review on using lazy-static**

I don't have a preference either way. This would be a better question for the rust-bitcoin people since the goal is for this lib to be under the rust bitcoin organization.

Probably a dumb question but could you just have it as a static and not a lazy-static? Evaluating it once probably doesn't hurt it too much and wouldn't introduce a new dependency.

Of course, if rust-bitcoin people are ok with it then I'm ok with it as well.

Davidson-Souza:
> Probably a dumb question but could you just have it as a static and not a lazy-static? Evaluating it once probably doesn't hurt it too much and wouldn't introduce a new dependency.  

AFAIK, you can't create a static var from a non-static function like `from_hex`, at least not in an out-of-the-box way. Static variables should be compilation-time know. You would have to implement lazy-static by yourself. I'm not aware of any way to create a non-zero hash that is const.
For instance, this wouldn't work:  
```rust
static ONE_AS_HASH: sha256::Hash = sha256::Hash::from_hex("0000000000000000000000000000000000000000000000000000000000000001").unwrap();
```

kcalvinalvin:
I played around with maybe changing the algorithm to not require the dependency but seems like there really isn't a clean way to do this.

If you rebase onto main, I'll merge it for now and worry about the extra dependency problem when we submit it to rust-bitcoin.

Davidson-Souza:
Rebased with main

kcalvinalvin:
ACK faeff4e25596727629a321a634dce4954e68e7f4

PR #22, opened 2023-01-11
Bump bitcoin hashes

Davidson-Souza:
Bump the version of bitcoin-hashes crate, and updates code
to conform with the new version. And remove duplicated dependency.
As explained in 4f6707d, depending on `bitcoin` and `bitcoin-hashes` is redundant, since we only need `bitcoin-hashes`.  

> We depend on bitcoin and bitcoin-hashes. Since we only need bitcoin::hashes from bitcoin, and bitcoin::hashes is a reexport of bitcoin-hashes, we don't need bitcoin. Furthermore, depending on bitcoin without specify features pulls libsecp and some other crates we don't use, but create problems for consumers if one of the libs try to use a different version of bitcoin. This commit drops it and use only bitcoin-hashes in its latest version.


kcalvinalvin:
ACK 4f6707dfc3554ca510d76274aee1304aaa312c6d

PR #23, opened 2023-01-11
feat: update_proof_del

Davidson-Souza:
This method is the first method to update cached proofs. You take a valid proof, and update with all deletion from a block, and get a proof valid for this new block, without needing to prove again.


kcalvinalvin:
Could you rebase on main? The change from `dafault` to `zero

Davidson-Souza:
Rebased

kcalvinalvin:
ACK ae74e64e5a238add561261a9961640ee46169320

Formatting looks good, verified that the small diffs are a result of branch `update-proof` not being rebased on main.

PR #26, opened 2023-01-25
Fix a compilation error in 1.41

### LINKED ISSUE #24: Can't use u64::BITS as it was added in version 1.53
https://github.com/mit-dci/rustreexo/blob/6a610fb860faf5a8cb33fc9e8b1b42f65276eeee/src/accumulator/util.rs#L279

The above line uses `u64::BITS` but we can't use it since the repo needs to support 1.41 but `u64::BITS` was added in version 1.53.

Also, odd that the CI is passing. The CI should've failed so something isn't quite working as it should

Davidson-Souza:
We use u64::BITS, but this was only added in 1.53. This PR uses the explicit constant instead. Fixes #24

### DIFF
--- a/src/accumulator/util.rs
+++ b/src/accumulator/util.rs
@@ -275,8 +275,9 @@ pub fn tree_rows(n: u64) -> u8 {
     if n == 0 {
         return 0;
     }
-
-    (u64::BITS - (n - 1).leading_zeros()) as u8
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
ACK d083cdce2f71e099836bd4fe1f760115bd1153b0

PR #27, opened 2023-02-04
Implement Pollard

Davidson-Souza:
Pollard is a data structure that holds prune-able trees to represent the current accumulator, thus, creating a client that may prove arbitrary UTXOs. This PR implements a Pollard struct in Rust with add/del methods.  

The implementation is capable of passing our test cases used for stump. There's also a wide range of tests for different aspects of a Pollard. 

All commits should compile and run on its own, modulo some warnings in the first ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,12 +22,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "112c678d4050afce233f4f2852bb2eb519230b3cf12f33585275537d7e41578d"
 
 [[package]]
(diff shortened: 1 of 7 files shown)


kcalvinalvin:
Is this ready for review?

Davidson-Souza:
I'll update this to support partial pollard. For full pollard, I made #36, which is ready to be reviewed and used in [here](https://github.com/Davidson-Souza/bridge). I could prove up to block 1M on testnet, but I ran out of RAM :/ (10GB isn't even close to be enough)

PR #32, opened 2023-04-24
Clean up the code

Davidson-Souza:
Remove unused files, and add crate and module-level docs.


src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
+//! This module is the core of the library. It contains all the basic data structures to use Utreexo,
+//! such as accumulator, and proofs. Even though the algorithms are equal, you may use different
```
kcalvinalvin, line 2: Maybe it's better to phrase it this way:

`s/Even though the algorithms are equal, you may use different data structures to represent the accumulator/The different data structures are compatible with each other and you may use different data structures to represent the accumulator./`

And then capitalize `Each` to make it a sentence.

src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
...
+//! data structures to represent the accumulator, each with its own advantages and disadvantages.
+//!
+//! # [Stump]
+//! A Stump is a basic data structure used in Utreexo. It only holds the roots and the number of leaves
+//! in the accumulator. This is useful to create lightweight nodes, the still validates, but is more compact,
```
kcalvinalvin: `s/This is useful to create lightweight nodes, the still validates, but is more compact, perfect to clients running on low-power devices./This is useful for creating lightweight nodes that still perform validation, making it suitable for clients on low-power devices.`

src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
...
+//! ```
+//!   use rustreexo::accumulator::{types::NodeHash, stump::Stump, proof::Proof};
+//!   use std::str::FromStr;
+//!   // Create a new empty Stump
+//!   let s = Stump::new();
+//!   // The newly create outputs
+//!   let utxos = vec![NodeHash::from_str("b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42").unwrap()];
+//!   // The spent outputs
+//!   let stxos = vec![];
+//!   // Modify the Stump, adding the new outputs and removing the spent ones, notice how
```
kcalvinalvin: `s/Modify the Stump, adding the new outputs and removing the spent ones,/Modify the Stump by adding new outputs and removing the spent ones./`

src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
...
+//!   // it returns a new Stump, instead of modifying the old one. This is due to the fact
+//!   // that modify is a pure function, and none of it's operands are modified in the process.
```
kcalvinalvin: Maybe state that the stump isn't modified/mutated rather than saying the operands are not modified since we don't care about the variables getting mutated.

src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
...
+//!   let s = s.modify(&utxos, &stxos, &Proof::default());
+//!   assert!(s.is_ok());
+//!   assert_eq!(s.unwrap().0.roots, utxos);
+//! ```
+//!
+//! # Proof
+//! A proof is a data structure that proves that a given element is in the accumulator. It is used
+//! to validate a transaction. A proof is composed of a list of hashes and a list of integers. The
+//! hashes are the siblings of the element in the accumulator. The integers are the position of the
```
kcalvinalvin: `s/hashes are the siblings of the element in the accumulator./hashes are the hashes need to calculate the root hash for validation./`

Something like that would be better since it can be the aunt of the element being proven.

src/accumulator/mod.rs
```diff
@@ -1,3 +1,67 @@
...
+//! element in the accumulator.
+//! ## Example
+//! ```
+//!   use bitcoin_hashes::{sha256, Hash, HashEngine};
+//!   use std::str::FromStr;
+//!   use rustreexo::accumulator::{types::NodeHash, stump::Stump, proof::Proof};
+//!   let s = Stump::new();
+//!   // Creates a tree with those values as leafs
+//!   let test_values:Vec<u8> = vec![0, 1, 2, 3, 4, 5, 6, 7];
+//!   // Targets are nodes witch we intend to prove
```
kcalvinalvin: `s/are nodes witch/are the nodes/`

src/lib.rs
```diff
@@ -1 +1,12 @@
+//! # Utreexo
+//! [Utreexo](https://eprint.iacr.org/2019/611.pdf) is a novel accumulator that allows for succinct
+//! UTXO set representation, using a logarithmic amount of space. It uses a dynamic accumulator
+//! that allows for the addition and deletion of elements. While spending  a UTXO, the element
```
kcalvinalvin, line 4: `s/While spending a  UTXO/When spending a UTXO/`

There's a double space here as well.

src/lib.rs
```diff
@@ -1 +1,12 @@
...
+//! is deleted from the accumulator. While receiving a UTXO, the element is added to the accumulator.
+//! During validation, nodes receive a proof that the UTXO they are spending is in the the current UTXO set
```
kcalvinalvin, line 6: `s/nodes receive a proof that the UTXO they are spending is in the the current UTXO set and has a given value./nodes receive a proof which proves the inputs in a block or transaction exist in the UTXO set./`

src/lib.rs
```diff
@@ -1 +1,12 @@
...
+//! and has a given value. Utreexo allows very succinct nodes, holding only a logarithmic amount of data.
+//!
+//! This lib have all basic building blocks to use Utreexo in a project. It is possible to create a
+//! new accumulator, add and delete elements, verify proofs, and serialize and deserialize then.
+//! FOr more information, check the documentation of each module.
```
kcalvinalvin, line 11: `s/FOr/For/`

src/accumulator/node_hash.rs
```diff
@@ -1,4 +1,38 @@
-// Rustreexo
+//! [NodeHash] is an internal type for representing Hashes in an utreexo accumulator. It's
+//! just a wrapper to [[u8; 32]] but with some useful methods.
```
kcalvinalvin, line 2: `s/just a wrapper to/just a wrapper around/`

src/accumulator/mod.rs
```diff
@@ -1,68 +1,14 @@
...
-//!   let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap().0;
-//!   // Create a proof for the targets
-//!   let p = Proof::new(targets, proof_hashes);
-//!   // Verify the proof
-//!   assert!(p.verify(&vec![hashes[0]] , &s).expect("This proof is valid"));
-//! ```
+//! All data structures implement the same high-level algorithms, and will always generate
+//! the same state.
+//!
+//! If you only need to verify proofs, and cache proofs used in your wallet only, see [Stump],
```
kcalvinalvin, line 7: `s/verify proofs, and cache proofs/verify proofs and cache proofs/`

src/accumulator/mod.rs
```diff
@@ -1,68 +1,14 @@
...
+//! a lightweight implementation of utreexo, that only stores roots. Although the [Stump]
```
kcalvinalvin, line 8: `s/of utreexo,/of utreexo/`

src/accumulator/proof.rs
```diff
@@ -1,6 +1,42 @@
+//! A proof is a data structure that proves that a given element is in the accumulator. It is used
```
kcalvinalvin, line 1: I'm going commit by commit. I think I addressed the comments here in a previous review comment

src/accumulator/stump.rs
```diff
@@ -1,4 +1,27 @@
-use super::{proof::Proof, types::NodeHash, util};
+//! A [Stump] is a basic data structure used in Utreexo. It only holds the roots and the number of leaves
```
kcalvinalvin, line 1: Here as well. I think I left a review comment about this in a different file.

kcalvinalvin (changes requested):
Mostly grammatical review. Code stuff looks ok. Good cleanup PR

PR #47, opened 2024-04-24
Fix pollard serialization/deserialization (#46), Debug/Display for empty pollard (#48)

### LINKED ISSUE #46: Serialization/Deserialization fails for empty node hashes
```
use rustreexo::accumulator::{node_hash::NodeHash, pollard::Pollard};

#[test]
fn test_pollard_serialization() {
    let mut p = Pollard::new();
    let values = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16];
    let hashes: Vec<NodeHash> = values
         .into_iter()
         .map(|i| NodeHash::from([i; 32]))
         .collect();
    p.modify(&hashes, &[]).expect("modify should work");
    p.modify(&[], &hashes).expect("modify should work");
    let mut serialized= Vec::<u8>::new();
    p.serialize(&mut serialized).expect("serialize should work");
    let _pollard = Pollard::deserialize(&*serialized).expect("deserialization fails");
}
```

Ash-L2L:
fixes #46, #48.
This fix is not backwards-compatible as the serialization encoding is changed. I have bumped the version number to reflect this.

### DIFF
--- a/src/accumulator/pollard.rs
+++ b/src/accumulator/pollard.rs
@@ -91,7 +91,7 @@ impl Node {
             NodeType::Branch => writer.write_all(&0_u64.to_le_bytes())?,
             NodeType::Leaf => writer.write_all(&1_u64.to_le_bytes())?,
         }
-        writer.write_all(&*self.data.get())?;
+        self.data.get().write(writer)?;
         self.left
             .borrow()
             .as_ref()
@@ -101,7 +101,7 @@ impl Node {
         self.right
             .borrow()
             .as_ref()
-            .map(|l| l.write_one(writer))
+            .map(|r| r.write_one(writer))
             .transpose()?;
         Ok(())
     }
@@ -118,10 +118,9 @@ impl Node {
             reader: &mut R,
             index: &mut HashMap<NodeHash, Weak<Node>>,
         ) -> std::io::Result<Rc<Node>> {
-            let mut data = [0u8; 32];
             let mut ty = [0u8; 8];
             reader.read_exact(&mut ty)?;
-            reader.read_exact(&mut data)?;
+            let data = NodeHash::read(reader)?;
 
             let ty = match u64::from_le_bytes(ty) {
                 0 => NodeType::Branch,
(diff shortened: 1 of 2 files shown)


Davidson-Souza:
Thank you for this PR! Just some comments

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rustreexo"
-version = "0.1.0"
+version = "0.2.0"
```
Davidson-Souza, line 3: Not sure if we should do this here. We probably need another release, but let's do it in another pr

src/accumulator/pollard.rs
```diff
@@ -622,6 +625,7 @@ mod test {
     use std::convert::TryFrom;
     use std::rc::Rc;
     use std::str::FromStr;
+    use std::task::Poll;
```
Davidson-Souza, line 628: I think this shouldn't be here.

src/accumulator/pollard.rs
```diff
@@ -972,4 +976,25 @@ mod test {
...
+        let hashes: Vec<NodeHash> = values
+            .into_iter()
+            .map(|i| NodeHash::from([i; 32]))
+            .collect();
+        p.modify(&hashes, &[]).expect("modify should work");
+        p.modify(&[], &hashes).expect("modify should work");
+        let mut serialized = Vec::<u8>::new();
+        p.serialize(&mut serialized).expect("serialize should work");
+        let _deserialized = Pollard::deserialize(&*serialized).expect("deserialize should work");
+    }
```
Davidson-Souza, line 999: Can we `assert` the roots here?

Davidson-Souza (changes requested):

src/accumulator/pollard.rs
Ash-L2L: Done, resolving

src/accumulator/pollard.rs
Ash-L2L: Removed, resolving

Davidson-Souza (approved):

PR #52, opened 2024-06-27
Improve calculate hashes

Davidson-Souza:
The current alg is okay for small proofs, but for big proofs, where we
need to move many elements to the left before adding to the middle of a
vector, this move becomes too expensive. The current alg uses two
vectors that are naturally sorted and we can just push to the end of one
vec.

This patch was tested on floresta and grants better performance while
not causing any incompatibility


Davidson-Souza:
Closing this in favor of #54

PR #57, opened 2024-10-08
feat: make the hash type generic

Davidson-Souza:
See https://github.com/keep-starknet-strange/raito/issues/248 for context

Currently, we only support using our internal hash type, that is the 32-bytes representation of a sha512-256 hash digest. However, we may want to give callers the possibility to use other hash types, or store them in other ways. This commit modifies `Pollard`, `Proof` and `Stump` to take in a generic parameter `Hash` that implements the trait `NodeHash` and defaults to `BitcoinNodeHash`, the one used by Bitcoin consensus as defined by the utreexo spec.

This is part of a project to support a Cairo prover for Bitcoin. To reduce the circuit size, we need to use an algebraic hash function like Poseidon, which tends to reduce the circuit significantly. With this commit, the caller can use our data structures with Poseidon without needing to change anything in rustreexo.

### DIFF
--- a/src/accumulator/node_hash.rs
+++ b/src/accumulator/node_hash.rs
@@ -1,14 +1,15 @@
-//! [NodeHash] is an internal type for representing Hashes in an utreexo accumulator. It's
+//! [AccumulatorHash] is an internal type for representing Hashes in an utreexo accumulator. It's
 //! just a wrapper around [[u8; 32]] but with some useful methods.
 //! # Examples
 //! Building from a str
 //! ```
 //! use std::str::FromStr;
 //!
-//! use rustreexo::accumulator::node_hash::NodeHash;
-//! let hash =
-//!     NodeHash::from_str("0000000000000000000000000000000000000000000000000000000000000000")
-//!         .unwrap();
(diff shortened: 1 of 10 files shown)


src/accumulator/node_hash.rs
```diff
@@ -58,39 +61,54 @@ use serde::Deserialize;
 #[cfg(feature = "with-serde")]
 use serde::Serialize;
 
+pub trait NodeHash: Copy + Clone + Ord + Debug + Display + std::hash::Hash + 'static {
```
m-kus, line 64: Just a thought regarding backward compatibility: would it make sense to keep the `NodeHash` struct as is and name the trait differently instead? I guess there are other projects (besides Floresta) depending on this crate and it would be a breaking change for them.

Davidson-Souza: Hmm, that's a good point. I wanted the struct to have a name that reflected the scope it had, so people wouldn't make confusion with the types. I guess I can keep the struct name, but rename `NodeHash` to `AccumulatorHash`. Then I could use a type alias `type NodeHash = BitcoinNodeHash` and deprecate it. That's backward-compatible, but sill would prevent confusion with the types.

PR #61, opened 2025-01-15
[bug fix] write roots len as 64-representation always

halseth:
On 32-bit platform usize would be written as 32-bit little endian, and
fail to be read during deserialization (was expecting 64-bit).


halseth:
@Davidson-Souza

PR #72, opened 2025-08-07
Serde: use hex encoding for BitcoinNodeHash when (de)serializer is human-readable

Ash-L2L:
This could be smaller with `hex-conservative::serde`, but unfortunately the `bitcoin-hashes/serde` feature does not enable `hex-conservative/serde`. Adding `hex-conservative` as a direct dependency in order to enable the feature would require manually matching the version of `hex-conservative` to the version used in `bitcoin-hashes`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        toolchain: [nightly, stable, 1.63.0]
+        toolchain: [nightly, stable, 1.73.0]
 
     runs-on: ${{ matrix.os }}
     steps:


Davidson-Souza:
So. With this, rather than a vector of integers like [242, 80, 99...] we would get a hexadecimal (`f25063...`). That's nice! I think it would be nice to have some tests, tho

Davidson-Souza:
Not sure about the MSRV bump, we follow [rust-bitcoin's](https://github.com/rust-bitcoin/rust-bitcoin?tab=readme-ov-file#minimum-supported-rust-version-msrv) for ecosystem compatibility.

Ash-L2L:
>Not sure about the MSRV bump, we follow [rust-bitcoin's](https://github.com/rust-bitcoin/rust-bitcoin?tab=readme-ov-file#minimum-supported-rust-version-msrv) for ecosystem compatibility.

Updated to only bump rust version for the CI tests. The version bump is only required for dev-dependencies, the crate will still build on `1.63.0`

PR #81, opened 2025-12-31
refactor!(stump): Move updated data to its own function

Davidson-Souza:
Before this change, `modify` would return the data needed to update a proof for the new block. This requires additional internal computation and extra allocations. During IBD we may never use this, since proof update is only meant for updating a few blocks worth of changes. Now there's a method specific to pull the modify_data, and `modify` itself will only return a new Stump.

When refactoring the addition function, I've modified it a little to allow a smarter one with a very nice property: it can pretend that it added a node, but actually represent it as deleted. The goal here is to do a Swift Sync-style protocol where you don't need deletions.

If a txout is already spent, you give an empty hash (BitcoinNodeHash::empty()). This will be exactly equivalent as giving this txo's hash and later on calling delete for with this txo's hash. However, it does this with only one call to modify and doesn't require proofs to achieve that.

This is an API breaking change, as modify now only returns one parameter, otherwise the behavior stays unchanged.

### DIFF
--- a/examples/proof_update.rs
+++ b/examples/proof_update.rs
@@ -22,7 +22,9 @@ fn main() {
     let utxos = get_utxo_hashes1();
     // Add the UTXOs to the accumulator. update_data is the data we need to update the proof
     // after the accumulator is updated.
-    let (s, update_data) = s.modify(&utxos, &[], &Proof::default()).unwrap();
+    let update_data = s.get_update_data(&utxos, &[], &Proof::default()).unwrap();
+    let s = s.modify(&utxos, &[], &Proof::default()).unwrap();
+
     // Create an empty proof, we'll update it to hold our UTXOs
     let p = Proof::default();
     // Update the proof with the UTXOs we added to the accumulator. This proof was initially empty,
@@ -48,8 +50,10 @@ fn main() {
     // We'll remove `0` as it got spent, and add 1..7 to our cache.
     let new_utxos = get_utxo_hashes2();
     // First, update the accumulator
-    let (stump, update_data) = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+    let stump = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+
     // and the proof
+    let update_data = s.get_update_data(&utxos, &[], &Proof::default()).unwrap();
     let (p2, cached_hashes) = p
         .update(
             cached_hashes,
--- a/examples/proof_update.rs
+++ b/examples/proof_update.rs
@@ -18,15 +18,20 @@ use rustreexo::stump::Stump;
 
 fn main() {
     let s = Stump::new();
+
     // Get the hashes of the UTXOs we want to insert
     let utxos = get_utxo_hashes1();
-    // Add the UTXOs to the accumulator. update_data is the data we need to update the proof
-    // after the accumulator is updated.
+
+    // Compute the update data for this block. Notice that we called this before updating the
+    // accumulator, as the accumulator state is required to compute this data.
     let update_data = s.get_update_data(&utxos, &[], &Proof::default()).unwrap();
+
+    // Now, update the accumulator with these UTXOs, no STXOs are being spent in this example.
     let s = s.modify(&utxos, &[], &Proof::default()).unwrap();
 
     // Create an empty proof, we'll update it to hold our UTXOs
     let p = Proof::default();
+
     // Update the proof with the UTXOs we added to the accumulator. This proof was initially empty,
     // but we can instruct this function to remember some UTXOs, given their positions in the list of
     // UTXOs we added to the accumulator. In this example, we ask it to cache 0 and 1.
@@ -37,6 +42,7 @@ fn main() {
     let (p, cached_hashes) = p
         .update(vec![], utxos.clone(), vec![], vec![0, 1], update_data)
         .unwrap();
+
     // This should be a valid proof over 0 and 1.
     assert_eq!(p.n_targets(), 2);
     assert_eq!(s.verify(&p, &cached_hashes), Ok(true));
@@ -49,11 +55,12 @@ fn main() {
     // Assume we have a block that (beyond coinbase) spends our UTXO `0` and creates 7 new UTXOs
     // We'll remove `0` as it got spent, and add 1..7 to our cache.
     let new_utxos = get_utxo_hashes2();
+
     // First, update the accumulator
     let stump = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
 
     // and the proof
-    let update_data = s.get_update_data(&utxos, &[], &Proof::default()).unwrap();
+    let update_data = s.get_update_data(&new_utxos, &[utxos[0]], &p1).unwrap();
     let (p2, cached_hashes) = p
         .update(
             cached_hashes,
--- a/src/proof/mod.rs
+++ b/src/proof/mod.rs
@@ -50,7 +50,7 @@
 //!     hashes.push(sha256::Hash::from_engine(engine).into())
 //! }
 //! // Add the UTXOs to the accumulator
-//! let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap().0;
+//! let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap();
 //! // Create a proof for the targets
 //! let p = Proof::new(targets, proof_hashes);
 //! // Verify the proof
@@ -129,7 +129,7 @@ pub(crate) type NodesAndRootsCurrent<Hash> = (Vec<(u64, Hash)>, Vec<Hash>);
 /// This is used when we need to return the nodes and roots for a proof
 /// if we are concerned with deleting those elements. The difference is that
 /// we need to retun the old and updatated roots in the accumulator.
-pub(crate) type NodesAndRootsOldNew<Hash> = (Vec<(u64, Hash)>, Vec<(Hash, Hash)>);
+pub(crate) type RootsOldNew<Hash> = Vec<(Hash, Hash)>;
 
 impl Proof {
     /// Creates a proof from a vector of target and hashes.
@@ -266,7 +266,7 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
     ///     engine.input(&[i]);
     ///     hashes.push(sha256::Hash::from_engine(engine).into())
     /// }
-    /// let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap().0;
+    /// let s = s.modify(&hashes, &vec![], &Proof::default()).unwrap();
     /// let p = Proof::new(targets, proof_hashes);
     /// assert!(s.verify(&p, &[hashes[0]]).expect("This proof is valid"));
     /// ```
@@ -439,7 +439,7 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
         &self,
         del_hashes: &[(Hash, Hash)],
         num_leaves: u64,
-    ) -> Result<NodesAndRootsOldNew<Hash>, String> {
+    ) -> Result<RootsOldNew<Hash>, String> {
         // Where all the root hashes that we've calculated will go to.
         let total_rows = util::tree_rows(num_leaves);
 
@@ -505,13 +505,7 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
             computed.push((parent, (old_parent_hash, parent_hash)));
         }
 
-        // we shouldn't return the hashes in the proof
-        nodes.extend(computed);
-        let nodes = nodes
-            .into_iter()
-            .map(|(pos, (_, new_hash))| (pos, new_hash))
-            .collect();
-        Ok((nodes, calculated_root_hashes))
+        Ok(calculated_root_hashes)
     }
 
     /// This function computes a set of roots from a proof.
@@ -581,7 +575,13 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
                 return Err(format!("Missing sibling for {next_pos}"));
             }
 
-            let parent_hash = AccumulatorHash::parent_hash(&next_hash, &sibling_hash);
+            let parent_hash = match (next_hash.is_empty(), sibling_hash.is_empty()) {
+                (true, true) => AccumulatorHash::empty(),
+                (true, false) => sibling_hash,
+                (false, true) => next_hash,
+                (false, false) => AccumulatorHash::parent_hash(&next_hash, &sibling_hash),
+            };
+
             let parent = util::parent(next_pos, total_rows);
             computed.push((parent, parent_hash));
         }
@@ -1021,7 +1021,10 @@ mod tests {
 
             let block_proof =
                 Proof::new(case_values.update.proof.targets.clone(), block_proof_hashes);
-            let (stump, updated) = stump.modify(&utxos, &del_hashes, &block_proof).unwrap();
+            let new_stump = stump.modify(&utxos, &del_hashes, &block_proof).unwrap();
+            let updated = stump
+                .get_update_data(&utxos, &del_hashes, &block_proof)
+                .unwrap();
             let (cached_proof, cached_hashes) = cached_proof
                 .update(
                     cached_hashes.clone(),
@@ -1032,7 +1035,7 @@ mod tests {
                 )
                 .unwrap();
 
-            let res = stump.verify(&cached_proof, &cached_hashes);
+            let res = new_stump.verify(&cached_proof, &cached_hashes);
 
             let expected_roots: Vec<_> = case_values
                 .expected_roots
@@ -1047,7 +1050,7 @@ mod tests {
                 .collect();
             assert_eq!(res, Ok(true));
             assert_eq!(cached_proof.targets, case_values.expected_targets);
-            assert_eq!(stump.roots, expected_roots);
+            assert_eq!(new_stump.roots, expected_roots);
             assert_eq!(cached_hashes, expected_cached_hashes);
         }
     }
@@ -1207,7 +1210,7 @@ mod tests {
     fn test_update_proof_delete() {
         let preimages = vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
         let hashes = preimages.into_iter().map(hash_from_u8).collect::<Vec<_>>();
-        let (stump, _) = Stump::new()
+        let stump = Stump::new()
             .modify(&hashes, &[], &Proof::default())
             .unwrap();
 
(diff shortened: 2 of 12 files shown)


src/accumulator/stump.rs
```diff
@@ -207,15 +220,39 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+    ///     BitcoinNodeHash::from_str(
+    ///         "b151a956139bb821d4effa34ea95c17560e0135d1e4661fc23cedc3af49dac42",
+    ///     )
+    ///     .unwrap(),
+    ///     BitcoinNodeHash::empty(), // This UTXO is going to be deleted in the future
+    /// ];
+    /// let stxos = vec![];
+    /// let s = s.modify(&utxos, &stxos, &Proof::default()).unwrap();
+    /// assert!(s.roots.contains(&utxos[0]));
+    /// assert!(!s.roots.contains(&utxos[1])); // The empty hash won't be in the roots
```
jaoleal: So if i got this correctly, you can pass a `BitcoinNodeHash::empty()` to tell the stump that this UTXO is spent in the future, making the acc not even adding it as a leave right ?

If im correct i would prefer to assert that the num leaves keeps being 1 and the roots num are 1.

correct me if im mistaken but, the roots in that case, if the acc included the other spent UTXO, would never be the spent UTXO,  so the check 
`/// assert!(!s.roots.contains(&utxos[1])); 
// The empty hash won't be in the roots` 
doesnt make sense ?

That feature is already proven when you assert that the first utxo became the root right ?

src/accumulator/stump.rs
```diff
@@ -191,7 +191,20 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
     /// matters, you can only modify, providing a list of utxos to be added,
     /// and txos to be removed, along with it's proof. Either may be
     /// empty.
+    ///
+    /// ## Adding stoxs to the stump
```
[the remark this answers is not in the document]
Davidson-Souza: No. Spent transaction outputs

Davidson-Souza:
@luisschwab 

>  @Davidson-Souza do you have a Floresta branch using this PR?

No, waiting on Jose's pr with swift sync stuff

JoseSK999:
It's time 🎅🏻

luisschwab:
nit: on the commit description: the `!` comes after the `(*)`, ie: `refactor(stump)!: ...`, per the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope-and--to-draw-attention-to-breaking-change) specification.

src/stump/mod.rs
```diff
@@ -207,15 +220,42 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+    /// assert!(!s.roots.contains(&utxos[1])); // The empty hash won't be in the roots
     /// ```
     pub fn modify(
         &self,
         utxos: &[Hash],
         del_hashes: &[Hash],
         proof: &Proof<Hash>,
-    ) -> Result<(Self, UpdateData<Hash>), StumpError> {
-        let (intermediate, mut computed_roots) = self.remove(del_hashes, proof)?;
+    ) -> Result<Self, StumpError> {
```
JoseSK999, line 282: Why no longer return `UpdateData`, a rebase issue? What is it for btw, I wonder

Davidson-Souza: > Before this change, modify would return the data needed to update a proof for the new block. This requires additional internal computation and extra allocations. During IBD we may never use this, since proof update is only meant for updating a few blocks worth of changes. Now there's a method specific to pull the modify_data, and modify itself will only return a new Stump.

luisschwab:
Needs rebase

Davidson-Souza:
Rebased

rustaceanrob:
Unable to offer conceptual review but code LGTM

examples/proof_update.rs
```diff
@@ -35,21 +42,26 @@ fn main() {
...
     // Assume we have a block that (beyond coinbase) spends our UTXO `0` and creates 7 new UTXOs
     // We'll remove `0` as it got spent, and add 1..7 to our cache.
     let new_utxos = get_utxo_hashes2();
+
     // First, update the accumulator
-    let (stump, update_data) = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+    let stump = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+
     // and the proof
+    let update_data = s.get_update_data(&utxos, &[], &Proof::default()).unwrap();
```
JoseSK999: Question: why aren't we passing here the deleted `utxos[0]` and the proof `p1`? Maybe a comment about it, it's confusing.

src/proof/mod.rs
```diff
@@ -471,15 +472,15 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
     /// It will compute all roots that contains elements in the proof, by hasing the nodes
     /// in the path to the root. This function returns the calculated roots and the hashes
     /// that were calculated in the process.
```
JoseSK999: This function no longer returns those hashes no? (forest nodes)

src/proof/mod.rs
```diff
@@ -561,7 +556,9 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
     /// hashes that were calculated in the process.
     /// This differs from `calculate_hashes_delelte` as this one is only used for verifying
     /// proofs, it doesn't compute the roots after the deletion, only the roots that are
-    /// needed for verification (i.e. the current accumulator).
+    /// needed for verification (i.e. the current accumulator). `calculate_hashes_delete` also
+    /// won't return the hashes that were calculated, we won't need them for updating the
```
JoseSK999: Nit, better?

```suggestion
    /// doesn't return the nodes that were calculated, as we won't need them for updating the
```

src/proof/mod.rs
```diff
@@ -1403,35 +1471,18 @@ mod tests {
...
 
-        let computed_positions = [0_u64, 1, 9, 13, 8, 12, 14].to_vec();
-        let computed_hashes = [
-            "0000000000000000000000000000000000000000000000000000000000000000",
-            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
-            "9576f4ade6e9bc3a6458b506ce3e4e890df29cb14cb5d3d887672aef55647a2b",
-            "29590a14c1b09384b94a2c0e94bf821ca75b62eacebc47893397ca88e3bbcbd7",
-            "4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a",
-            "2b77298feac78ab51bc5079099a074c6d789bd350442f5079fcba2b3402694e5",
-            "726fdd3b432cc59e68487d126e70f0db74a236267f8daeae30b31839a4e7ebed",
```
JoseSK999, line 1453: Only a question, is there a different place where we could repurpose these constants to check the computed hashes?

src/stump/mod.rs
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! A [Stump] is a basic data structure used in Utreexo. It only holds the roots and the number of leaves
-//! in the accumulator. This is useful to create lightweight nodes, the still validates, but is more compact,
-//! perfect to clients running on low-power devices.
+//! in the accumulator. This is useful to create lightweight nodes, the still validates, but is more compact.
+//! This is useful for clients running is super constrained environments, such as mobile devices.
```
JoseSK999: Nit: you duplicate the "this is useful to/for" line, you could make this a single phrase, without the grammar errors

src/stump/mod.rs
```diff
@@ -206,10 +206,23 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
     }
 
     /// Modify is the external API to change the accumulator state. Since order
-    /// matters, you can only modify, providing a list of utxos to be added,
-    /// and txos to be removed, along with it's proof. Either may be
+    /// matters, you can only modify, providing a list of UTXOs to be added,
+    /// and TXOs to be removed, along with it's proof. Either may be
```
JoseSK999: Nit: makes it clear that the proof is for the latter

```suggestion
    /// matters, you can only modify providing a list of UTXOs to be added, and
    /// the TXOs to be removed along with their inclusion proof. Either may be
```

src/stump/mod.rs
```diff
@@ -206,10 +206,23 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
     /// empty.
+    ///
+    /// ## Adding STXOs to the stump
```
JoseSK999:
```suggestion
    /// ## Adding spent TXOs (STXOs) to the stump
```

src/stump/mod.rs
```diff
@@ -206,10 +206,23 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+    ///
+    /// If you know that an element will be deleted in the future, you can
+    /// pass an empty hash as the UTXO to be added. There's a possible
+    /// optimization here, where you can apply the changes caused by deletion
+    /// during addition. If you do so, you don't need to call `modify` when
+    /// this leaf gets deleted. This will save you some hashing, as well as
+    /// you won't need proofs anymore for those leaves, thus saving you
+    /// bandwidth.
```
JoseSK999: A bit more explanatory, mentioning the SwiftSync case where this can be used.

I think we also need to explain that the accumulator is not valid until we reach the final state transition where implicit deletions refer to (e.g., the SS stop height). Before that point the SS accumulator will be different from the sequential one as we remove elements ahead of time.

```
    /// If you know that an element will be deleted in the future, as is the case
    /// in SwiftSync, you can pass an empty hash as the UTXO to be added.
    ///
    /// This is an optimized version of deletion that we call "implicit deletion",
    /// as you apply the changes caused by deletion during addition. If you do so,
    /// you don't need to later call `modify` to delete this leaf (it was already
    /// deleted just after being added). This will save you some hashing, as well
    /// as the need to download proofs for those leaves, thus saving you bandwidth.
```

src/stump/mod.rs
```diff
@@ -206,10 +206,23 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+    ///
     ///# Example
```
JoseSK999:
```suggestion
    /// # Example
    ///
```

src/stump/mod.rs
```diff
@@ -206,10 +206,23 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+    /// Using the normal addition
```
JoseSK999: Maybe?


```suggestion
    /// ###### Using the normal addition
```

src/stump/mod.rs
```diff
@@ -225,15 +238,42 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
     /// let stxos = vec![];
     /// let s = s.modify(&utxos, &stxos, &Proof::default());
     /// assert!(s.is_ok());
-    /// assert_eq!(s.unwrap().0.roots, utxos);
+    /// assert_eq!(s.unwrap().roots, utxos);
+    /// ```
+    ///
+    /// Using addition with implicit deletion
```
JoseSK999: Maybe^2?

```suggestion
    /// ###### Using addition with implicit deletion
```

JoseSK999 (commented):
More nits

examples/proof_update.rs
JoseSK999: Everywhere else you pass the delete hashes and proof in `get_update_data`, so it seems wrong

Davidson-Souza:
Applied docs suggestions by @JoseSK999

src/proof/mod.rs
Davidson-Souza: We use similar tests in several other places, I don't think removing this here would be any problem.

JoseSK999:
First 3 comments not addressed yet

Davidson-Souza:
> First 3 comments not addressed yet

Shoot, sorry! Should be fixed now

src/stump/mod.rs
```diff
@@ -254,21 +301,112 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
 
         let new_stump = Self {
             leaves: self.leaves + utxos.len() as u64,
             roots,
         };
 
+        Ok(new_stump)
+    }
+
+    /// Return the data needed to update a proof with the changes made a block.
```
lorenzolfm: made in a block

src/stump/mod.rs
```diff
@@ -331,8 +466,32 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
             .map_err(StumpError::InvalidProof)
     }
 
+    fn add(mut roots: Vec<Hash>, utxos: &[Hash], mut leaves: u64) -> Vec<Hash> {
+        for add in utxos.iter() {
+            let pos = leaves;
+            let mut h = 0;
+            let mut to_add = *add;
+            while (pos >> h) & 1 == 1 {
+                let root = roots.pop().unwrap();
```
lorenzolfm: expect?

src/proof/mod.rs
```diff
@@ -484,17 +485,16 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
     /// This function computes a set of roots from the proof
     ///
     /// It will compute all roots that contains elements in the proof, by hasing the nodes
```
JoseSK999:
```suggestion
    /// It will compute all roots that contain elements in the proof, by hashing the nodes
```

src/proof/mod.rs
```diff
@@ -583,7 +577,9 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
     /// hashes that were calculated in the process.
     /// This differs from `calculate_hashes_delelte` as this one is only used for verifying
     /// proofs, it doesn't compute the roots after the deletion, only the roots that are
-    /// needed for verification (i.e. the current accumulator).
+    /// needed for verification (i.e. the current accumulator). `calculate_hashes_delete` also
+    /// doesn't return hashes that were calculated, as we won't need them for updating the
```
JoseSK999:
```suggestion
    /// doesn't return nodes that were calculated, as we won't need them for updating the
```

src/stump/mod.rs
```diff
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! A [Stump] is a basic data structure used in Utreexo. It only holds the roots and the number of leaves
-//! in the accumulator. This is useful to create lightweight nodes, the still validates, but is more compact,
-//! perfect to clients running on low-power devices.
+//! in the accumulator. This is useful to create lightweight nodes, the still validates, but is more compact.
```
JoseSK999:
```suggestion
//! in the accumulator. This is useful to create lightweight nodes which still validate, but is more compact.
```

src/stump/mod.rs
```diff
@@ -324,8 +459,34 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
             .map_err(StumpError::InvalidProof)
     }
 
+    fn add(mut roots: Vec<Hash>, utxos: &[Hash], mut leaves: u64) -> Vec<Hash> {
+        for add in utxos.iter() {
+            let pos = leaves;
+            let mut h = 0;
+            let mut to_add = *add;
+            while (pos >> h) & 1 == 1 {
```
JoseSK999, line 475: Add a comment to explain what the bitwise op means because I forgot your explanation already

src/stump/mod.rs
```diff
@@ -324,8 +459,34 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+            }
+            roots.push(to_add);
+            leaves += 1;
+        }
+
+        roots
+    }
+
     /// Adds new leaves into the root
-    fn add(
```
JoseSK999: You should move the docstring to the new `add` method since you renamed this to `compute_update_data_add`, and use a different docstring for this

src/stump/mod.rs
JoseSK999: We need a type alias for this if we don't have it already

Davidson-Souza:
Applied nits by @JoseSK999

src/stump/mod.rs
```diff
@@ -247,21 +299,112 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+        let mut new_roots = vec![];
+
+        for root in self.roots.iter() {
+            if let Some(pos) = computed_roots.iter().position(|(old, _new)| old == root) {
+                let (old_root, new_root) = computed_roots.remove(pos);
+                if old_root == *root {
+                    new_roots.push(new_root);
+                    continue;
+                }
+            }
```
JoseSK999, line 392: The `if` clause is always true, we can do this (same code as in `modify`)

```rust
if let Some(pos) = computed_roots.iter().position(|(old, _new)| old == root) {
    let (_, new_root) = computed_roots.remove(pos);
    new_roots.push(new_root);
    continue;
}
```

src/stump/mod.rs
JoseSK999, line 536: Shouldn't we use the match arm here to also account for empty hashes?

src/stump/mod.rs
```diff
@@ -247,21 +299,112 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
...
+
+            new_roots.push(*root);
+        }
+        let (_, updated, destroyed) = Self::compute_update_data_add(new_roots, utxos, self.leaves);
```
JoseSK999: Shouldn't we also do this here?

```rust
        // If there are still roots to be added, it means that the proof is invalid
        // as we should have consumed all the roots.
        if !computed_roots.is_empty() {
            return Err(StumpError::InvalidProof(ProofError::RootsMismatch));
        }
        
        let (_, updated, destroyed) = ...
```

examples/proof_update.rs
```diff
@@ -47,9 +55,12 @@ fn main() {
     // Assume we have a block that (beyond coinbase) spends our UTXO `0` and creates 7 new UTXOs
     // We'll remove `0` as it got spent, and add 1..7 to our cache.
     let new_utxos = get_utxo_hashes2();
+
     // First, update the accumulator
-    let (stump, update_data) = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+    let stump = s.modify(&new_utxos, &[utxos[0]], &p1).unwrap();
+
     // and the proof
+    let update_data = s.get_update_data(&utxos, &[utxos[0]], &p1).unwrap();
```
JoseSK999: This was wrong

```suggestion
    let update_data = s.get_update_data(&new_utxos, &[utxos[0]], &p1).unwrap();
```

src/stump/mod.rs
```diff
@@ -38,11 +37,16 @@ use serde::Serialize;
...
-use super::proof::NodesAndRootsOldNew;
 use super::proof::Proof;
 use super::proof::ProofError;
 use super::util;
 use crate::prelude::*;
+use crate::proof::RootsOldNew;
+
+/// A type representing what's changed for addition.
+///
+/// It returns the final roots destroyed hashes and  all the new positions created during this add.
```
JoseSK999: Nit: I think order is inaccurate and there's a double space

JoseSK999 (changes requested):

src/stump/mod.rs
Davidson-Souza: This function shouldn't use implicit deletions as we need the proof hashes to update the proof anyway.

src/stump/mod.rs
Davidson-Souza: `get_update_data` assumes you already validated that block, but I can add it out of caution.

Davidson-Souza:
@JoseSK999 done!

JoseSK999 (approved):
ACK e1afeb7; tested this commit with SwiftSync in Floresta, it synced and continued with Utreexo proofs

PR #88, opened 2026-01-19
feat: add height and hash to the `Stump`

### LINKED ISSUE #80: Come up with a de/serialization standard for the Accumulator State

luisschwab:
This PR adds the height and hash to the `Stump`, so we can just de/serialize it to transfer validation across devices.

Maybe closes #80


luisschwab:
Perhaps adding a wrapper type on floresta with the `Stump` and these extra fields?

PR #93, opened 2026-02-05
feat: enable `no_std` compilation and check it in CI

JoseSK999:
### Description

This PR adds two new dependencies that are needed for `no_std` support: `hashbrown` (`HashMap` alternative) and `bitcoin_io` (`io` trait and type replacements).

The `AccumulatorHash` trait bounds depend on whether we use `std` or `no_std` now.

### Note

This seems to work but perhaps we should add a feature powerset CI check to ensure all 4 combinations work fine for clippy and testing.

The `prelude` list is from `floresta-common`, I think I can trim a couple of these imports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,16 @@ assert!(stump.verify(utxo_proof).unwrap());
 
 for a complete example, see `examples/`.
 
+## Minimum Supported Rust Version (MSRV)
+
+This library should compile with any combination of features on Rust **1.74.0**.
+
+## No-std support
+
+The `std` cargo feature is enabled by default. To build this project without the Rust standard
+library, use the `--no-default-features` flag or set `default-features = false` in your dependency
+declaration when adding it to your project.
+
 ## Testing
 
 This library contains an exhaustive test suite that covers all the algorithms implemented.


src/accumulator/mem_forest.rs
```diff
@@ -999,11 +995,10 @@ mod test {
         let mut p = MemForest::new();
         p.modify(&hashes, &[]).expect("Test mem_forests are valid");
         p.modify(&[], &[hashes[0]]).expect("can remove 0");
-        let mut writer = std::io::Cursor::new(Vec::new());
+        let mut writer = io::Cursor::new(Vec::new());
```
Davidson-Souza: nit: import `Cursor`

src/accumulator/pollard.rs
```diff
@@ -1095,7 +1097,7 @@ impl<Hash: AccumulatorHash> Pollard<Hash> {
         let mut roots_to_destroy = Vec::new();
 
         while self.leaves >> row & 1 == 1 {
-            let old_root = std::mem::take(&mut self.roots[row as usize]).expect("Root not found");
+            let old_root = core::mem::take(&mut self.roots[row as usize]).expect("Root not found");
```
Davidson-Souza: nit: import `take`

src/accumulator/pollard.rs
```diff
@@ -493,11 +495,11 @@ impl<Hash: AccumulatorHash> PollardNode<Hash> {
     /// node's children. This function swaps the nieces of this node with the nieces of the provided
     /// node.
     fn swap_nieces(&self, other: &Self) {
-        std::mem::swap(
+        core::mem::swap(
```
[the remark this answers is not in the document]
JoseSK999: In this case most people at least specify `mem::swap` no?

JoseSK999:
> @JoseSK999 you could also add a `test-no-std` recipe to the `justfile` and have CI run the recipe, so we can avoid duplication and have the `justfile` as the single source of truth for testing.

I had extended the `just test` recipe to test no default features (i.e., `no-std`), I think better than a separate `just test-no-std`. But CI is not using `just test`. Should I make the recipe be

```justfile
test:
    cargo +nightly test --no-default-features
    cargo +nightly test --all-features
    cargo +stable test --no-default-features
    cargo +stable test --all-features
```

and use it in CI? (not sure)

src/accumulator/pollard.rs
```diff
@@ -1,47 +1,50 @@
...
+//! also serve proofs for specific transactions as requested, allowing efficient transaction relay.
+//!
+//! This implementation is close to the one in `MemForest`, but it is specialized in keeping track
+//! of subsets of the whole tree, allowing you to cache and uncache elements as needed. While the
+//! MemForest keeps everything in the accumulator, and may take a lot of memory.
+//!
+//! Nodes are kept in memory, and they hold their hashes, a reference to their **aunt** (not
+//! parent!), and their nieces (not children!). We do this to allow for proof generation, while
+//! prunning as much as possible. In a merkle proof, we only need the sibling of the path to the
+//! root, the parent is always computed on the fly as we walk up the tree. Some there's no need to
```
Davidson-Souza:
```suggestion
//! root, the parent is always computed on the fly as we walk up the tree. So, there's no need to
```

JoseSK999: Done

luisschwab (changes requested):

luisschwab (approved):
ACK e93a3dfe54377ef04da1ffdf4430276d0a1c68a5

Davidson-Souza:
@JoseSK999 could you mention on readme that we support `no_std` by default? 

I think an embed example (with `#[no_main]`) would be cool as well, something like [this](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/embedded/src/main.rs).  However, it might be better as a follow up

PR #124, opened 2026-05-12
Bump RBMT to v0.2.1

luisschwab:
Overrides #123 

## Changelog
```
- Bump RBMT to v0.2.1:
  - Add a vendored `setup-rbmt` action
- Update the `justfile`
- Fix doc comment references (`rbmt` v0.2.1 now checks examples)
```

### DIFF
--- a/.github/actions/setup-rbmt/action.yml
+++ b/.github/actions/setup-rbmt/action.yml
@@ -1,0 +1,31 @@
+name: Setup cargo-rbmt
+
+description: Install cargo-rbmt
+
+inputs:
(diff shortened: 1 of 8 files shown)


luisschwab:
I guess nightly toolchain + minimal deps is too much. Actually, the `nightly` toolchain should not be on the test matrix, as nightly should be used for formatting only.

luisschwab:
I'm doing/reviewing some stuff on `cargo-rbmt` and will update this after that.

luisschwab:
Replaced by #126

PR #129, opened 2026-07-02
refactor: improve some error messages for Stump

Davidson-Souza:
We currently use some messages that are hard to debug, or outright wrong. For example, if a proof is invalid because we didn't compute the original rootset, we return `EmptyProof`, which is misleading.

This commit adds a few more granualar errors and use where appropriate.

### DIFF
--- a/src/stump/mod.rs
+++ b/src/stump/mod.rs
@@ -59,10 +59,6 @@ pub struct UpdateData<Hash: AccumulatorHash> {
 #[derive(Clone, Debug, PartialEq, Eq)]
 /// The error kinds returned by Stump's methods
 pub enum StumpError {
-    /// User passed in a proof that doesn't have any elements, but a non-zero list of
-    /// targets.
-    EmptyProof,
-
     /// An IO error occurred, this is usually due to a failure in reading or writing
     /// the Stump to a reader/writer. This error will be returned during (de)serialization.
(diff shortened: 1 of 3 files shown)


luisschwab (approved):
ACK 21788d4cc34ced7b3e6483e07b89015611276dae

src/stump/mod.rs
JoseSK999: Unused now no?

src/stump/mod.rs
```diff
@@ -238,11 +238,9 @@ impl<Hash: AccumulatorHash> Stump<Hash> {
 
         for root in self.roots.iter() {
             if let Some(pos) = computed_roots.iter().position(|(old, _new)| old == root) {
-                let (old_root, new_root) = computed_roots.remove(pos);
-                if old_root == *root {
```
JoseSK999, line 242: It was redundant I see

JoseSK999 (approved):
ACK 21788d4

lorenzolfm (approved):
ACK

Davidson-Souza:
Removed an unused error variant in `StumpError`

luisschwab (approved):
re-ACK 0b3c3a7e6f1dcc1000b50b7dd0eac17815191e0f

I'll add `rust-bitcoin`'s lints here as well; they would've caught the unused error.

PR #140, opened 2026-07-08
Add and apply most of `rust-bitcoin`'s lints

luisschwab:
## Changelog
```
- Bump `edition` to 2021 (supported since Rust 1.56.0)
- Pins `cargo-audit` and `zizmor` versions on `Cargo.toml`, such that local and CI use the same versions
- Implement `new_hash_map` and `new_hash_set` for the `std` prelude
- Add and apply all of `rust-bitcoin`'s lints (except for documentation lints, left to a follow-up)
- Add the `#![no_std]` directive such that `cargo-rbmt` will also test on a no-std target
- Bump toolchains and fix `clippy::mut_mut` lint on `Stump::serialize`
```

PR #141, opened 2026-07-08
fix: use correct cargo category

Davidson-Souza:
crates.io is complaining about cryptography::Cryptocurrencies because it is case-sensitive, and currently the 'C' in 'Cryptocurrencies' is captilized where it shouldn't.

It also bumps version for release.


luisschwab:
Why bump to 0.6.1?

Davidson-Souza:
> Why bump to 0.6.1?

It's a very minor fix since `0.6.0`, I thought it wouldn't be enough for a `0.7.0`

luisschwab:
This is doc-only. I don't think it warrants a new version, unless you want to merge #140 before v0.6.1. Then it makes sense.

Davidson-Souza:
> This is doc-only. I don't think it warrants a new version, unless you want to merge #140 before v0.6.1. Then it makes sense.

It's weird to have a release in a commit that's not version bump. 

> unless you want to merge https://github.com/mit-dci/rustreexo/pull/140 before v0.6.1

I think we should leave #140 to the next cycle

luisschwab:
> It's weird to have a release in a commit that's not version bump.

It's fine, you can just tag this merge commit with v0.6.0

Davidson-Souza:
@luisschwab dropped [4b84d48](https://github.com/mit-dci/rustreexo/commit/4b84d4801fb5bed6c99be79d8406387b85a40532)

luisschwab (approved):
ACK b03204058cd7fcb0964bb7e1a87d2f4245ed6930

PR #143, opened 2026-07-25
proof: Bound deserialize against wire-length OOM

nervana21:
### Description and Notes

`Proof::deserialize` previously used untrusted `u64` length prefixes to size `Vec::with_capacity` for targets and hashes. A hostile payload could claim a huge count and OOM the process before any proof bytes were validated.

This PR rejects oversized prefixes before allocation. Target and hash counts above `MAX_PROOF_DESERIALIZE_COUNT` (`(4 * 1024³) / 32 == 1 << 27`, ~4 GiB of 32-byte hashes) return `ProofError::OversizedAllocation`. Cap is deserialize-only. `Proof::serialize` and in-memory construction are unchanged.

#### Changelog

```
proof: Bound deserialize against wire-length OOM
 - reject target/hash counts above MAX_PROOF_DESERIALIZE_COUNT before allocating
 - return ProofError::OversizedAllocation for oversized length prefixes
 - add unit tests for excessive target and hash caps
```

### To verify changes

- `cargo test test_deserialize_rejects`
- `cargo test --lib`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,17 @@ extern crate alloc;
 /// That should save you the trouble.
 pub(crate) const MAX_FOREST_ROWS: u8 = 63;
 
+/// Maximum number of targets **or** proof hashes accepted by
+/// [`proof::Proof::deserialize`].
+///
+/// Untrusted length prefixes are checked against this value before
+/// `Vec::with_capacity`, so a hostile payload cannot force an enormous
+/// reservation and OOM the process. The bound is intentionally large and
+/// fixed (~4 GiB of 32-byte hashes: `(4 * 1024³) / 32 == 1 << 27`).
+/// The same count applies to targets (`u64`). [`proof::Proof::serialize`]
+/// and in-memory construction are uncapped.
+pub const MAX_PROOF_DESERIALIZE_COUNT: u64 = (4u64 * 1024 * 1024 * 1024) / 32;
+
 #[cfg(not(feature = "std"))]
 /// Re-exports `alloc` basics plus HashMap/HashSet and IO traits.
 pub mod prelude {
--- a/src/proof/mod.rs
+++ b/src/proof/mod.rs
@@ -73,11 +73,13 @@ use super::node_hash::BitcoinNodeHash;
 use super::stump::UpdateData;
 use super::util;
 use super::util::get_proof_positions;
+use super::util::read_bounded_len;
 use super::util::read_u64;
 use super::util::tree_rows;
 use crate::prelude::*;
 use crate::util::translate;
 use crate::MAX_FOREST_ROWS;
+use crate::MAX_PROOF_DESERIALIZE_COUNT;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 /// Errors that can occur when working with a [Proof].
@@ -90,6 +92,9 @@ pub enum ProofError {
 
     /// A hash could not be parsed during deserialization.
     InvalidHash,
+
+    /// A length prefix claimed more elements than allowed during deserialization.
+    OversizedAllocation { requested: u64, max: u64 },
 
     /// The computed roots don't match current accumulator.
     RootsMismatch,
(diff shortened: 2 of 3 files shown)


src/proof/mod.rs
```diff
@@ -133,6 +142,15 @@ impl From<io::Error> for ProofError {
...
 
+/// Reads a little-endian `u64` length and ensures it is less than or equal to `max`.
+fn read_bounded_len<R: Read>(reader: &mut R, max: u64) -> Result<usize, ProofError> {
+    let n = read_u64(reader).map_err(|e| ProofError::Io(e.kind()))?;
+    if n > max {
+        return Err(ProofError::OversizedAllocation { requested: n, max });
+    }
+    Ok(n as usize)
+}
+
```
Davidson-Souza: Don't we have an `util` for this? If not, this could live there

src/lib.rs
```diff
@@ -53,6 +53,15 @@ extern crate alloc;
 /// That should save you the trouble.
 pub(crate) const MAX_FOREST_ROWS: u8 = 63;
 
+/// Deserialize DoS cap on proof hashes (~32 MiB at 32 B/hash).
+pub const MAX_PROOF_HASH_COUNT: u64 = 1 << 20;
+
+/// Deserialize DoS cap on target count.
+pub const MAX_PROOF_TARGET_COUNT: u64 = 1 << 20;
```
Davidson-Souza: I think the rational for these should be better explained

src/lib.rs
```diff
@@ -53,6 +53,15 @@ extern crate alloc;
...
+
+/// Max rootward path length (`MAX_FOREST_ROWS + 1`). Caps hashes per target.
```
Davidson-Souza: Why the `+1`?

src/proof/mod.rs
```diff
@@ -459,14 +491,19 @@ impl<Hash: AccumulatorHash> Proof<Hash> {
...
         let mut targets = Vec::with_capacity(targets_len);
         for _ in 0..targets_len {
             targets.push(read_u64(&mut buf).map_err(|_| ProofError::InvalidTarget)?);
         }
 
-        let hashes_len = read_u64(&mut buf).map_err(|e| ProofError::Io(e.kind()))? as usize;
+        // Valid proofs need at most one sibling hash per target per tree level.
+        // Also apply an absolute DoS cap so large target counts cannot reserve
+        // targets_len * depth * 32 bytes.
+        let max_hashes = (targets_len as u64)
```
Davidson-Souza: Use explicit conversion

Davidson-Souza (changes requested):

src/proof/mod.rs
nervana21: Done.

src/proof/mod.rs
nervana21: Done.

src/lib.rs
nervana21: I've updated the comments to better explain the rational. Please let me know what you think.

src/lib.rs
nervana21: I think you're right and that I was OBO. I've updated the code to directly use the `MAX_FOREST_ROWS` constant instead.

Davidson-Souza:
I thought about this a little more. Ideally this library won't use Bitcoin specific values — we do have other consumers that aren't doing the default Bitcoin stuff we do in Floresta.

Floresta doesn't even use this deserializer internally. I like the patch, but I would just pick a big number — like 4GB — and stick with it, rather than computing Bitcoin-specific upper bounds.

Edit: and probably make this explicit in the docs.

nervana21:
_**Updates**_:

* Addressed @Davidson-Souza's [feedback](https://github.com/mit-dci/rustreexo/pull/143#issuecomment-5220343488).

src/lib.rs
```diff
@@ -53,6 +53,17 @@ extern crate alloc;
...
 pub(crate) const MAX_FOREST_ROWS: u8 = 63;
 
+/// Maximum number of targets **or** proof hashes accepted by
+/// [`proof::Proof::deserialize`].
+///
+/// Untrusted length prefixes are checked against this value before
+/// `Vec::with_capacity`, so a hostile payload cannot force an enormous
+/// reservation and OOM the process. The bound is intentionally large and
+/// fixed (~4 GiB of 32-byte hashes: `(4 * 1024³) / 32 == 1 << 27`).
+/// The same    count applies to targets (`u64`). [`proof::Proof::serialize`]
```
Davidson-Souza: Nit
```suggestion
/// The same count applies to targets (`u64`). [`proof::Proof::serialize`]
```

Davidson-Souza (changes requested):
Last nit

src/lib.rs
nervana21: Done!

luisschwab (approved):
ACK c06d8524454f1ba0ec05e8da8688ebe2ce6a2c28

Davidson-Souza (approved):
ACK c06d8524454f1ba0ec05e8da8688ebe2ce6a2c28

PR #149, opened 2026-08-06
fuzz: add deserialize, proof_corruption, stump_model harness

brunoerg:


Davidson-Souza:
Merged as part of #154

PR #92, opened 2026-02-03
feat: add fuzz integration

Davidson-Souza:
This PR adds some basic fuzz integration, with the following targets:
 - `pollard_ser`: Deserializes the random data as a pollard, serialize it deserialize again and assert they are equal
 - `pollard_operations`: Creates a few  mock block, then applies the deltas to those blocks and checks whether the operations works

PR #154, opened 2026-08-26
feat: add fuzz integration [take two]

Davidson-Souza:
Partially replaces #92
Replaces #149 

This pull request adds fuzzing capabilities to `rustreexo`, including  targets to test our cryptographic code, focusing on `stump` — since this is the implementation we currently use in floresta.

The first target tests proof and stump [de]serialization, making sure it doesn't panic even on malformed inputs. The second creates intentionally invalid proofs and checks how our stump behaves.

These targets were initially written by Bruno in #149, I just cleaned them up and fixed problems found while running them.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,140 @@
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+name: Fuzz Tests
+
+on:
+  push:
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_FUZZ_VERSION: 0.13.1
+  APT_CONFIG: |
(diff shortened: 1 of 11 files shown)


luisschwab:
Could add a fuzz CI job that runs $n$ iterations, just to verify the fuzzers work, and a `justfile` recipe so we can fuzz locally.

Davidson-Souza:
> Could add a fuzz CI job that runs n iterations, just to verify the fuzzers work, and a `justfile` recipe so we can fuzz locally.

@luisschwab done!

Davidson-Souza:
Added another regression test.

luisschwab (approved):
ACK 0eccb20378e9b454f118bdc181de510b38822c07

